PR #2, opened 2014-02-18
Fix for Ticket #7729

ahmedcharles:
concept_def.hpp and concept_undef.hpp need to be matched when used.


vprus:
Thanks for the patch, it has been cherry-picked.

PR #16, opened 2014-08-01
Fixed bug 10231 partly: If finish_edge was called, then now correctly.

gatlex:
Written in conjunctive because the bug that it never gets called with the current construction remains.


gopchandani:
I am observing this behavior in boost version 1.67... I tried the test file, same thing, the finish_edge in the visitor does not get called, whereas I can see others being called...

My g++/gcc version is 5.4.0.

PR #27, opened 2014-12-18
VF2, fix assumption that index maps are default constructible.

jakobandersen:
The fix moves construction of some internal data structures into the relevant initialization list. The added test case triggers the issue in the previous revisions.


murraycu:
This does look much cleaner anyway.

Maybe that test should have a comment saying what it is testing.

(I'm not the maintainer.)

PR #30, opened 2015-01-23
Remove deprecated type traits idiums

jzmaddock:
The undocumented ice_* types are being deprecated - no current compilers appear to require them, this fix removes their usage from this header.


Belcourt:
John, can we post this PR against develop and close this one against master.  The Boost development model et. al.  Noel.

jzmaddock:
Apologies, my bad.

PR #32, opened 2015-02-12
Make it clearer that when tree_edge is called

xuhdev:
tree_edge function is called before the target vertex is discovered, so users should be acknoledged that at this moment, the target's color is still white.

---

There may be a more proper way to fix the document, but I think the original version is confusing, as the target is actually still white.


Belcourt:
The Boost development model is to post PR's against the develop branch.  We only apply changes to master from the develop branch.  Please close this PR against master and submit against develop.

xuhdev:
OK, new PR made. Thanks

PR #34, opened 2015-03-21
Fix Ticket #11133: Graph: using 'nil' as a local variable

jaredgrubb:
This is effectively a cosmetic change, as it only changes the name of a local variable. However, when this header gets in included in code compiled under ObjC++, the compiler doesnt like seeing 'nil' as a local variable (it's a keyword).

I wasnt sure what name to pick, so I picked "nil_node". I'm happy to adjust it to something else.

### DIFF
--- a/include/boost/graph/detail/list_base.hpp
+++ b/include/boost/graph/detail/list_base.hpp
@@ -58,10 +58,10 @@ namespace boost {
 
     template <class Node, class Next>
     inline Node
-    slist_previous(Node head, Node x, Node nil, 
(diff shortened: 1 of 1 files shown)


Belcourt:
nil_node is okay, but I think it'd read better if it was changed to 'empty'.  Does that seem reasonable to you?

jaredgrubb:
I totally missed this comment. I have adjusted the variable name to "empty" and rebased the commit.

jaredgrubb:
Any hope for merging this soon?

jaredgrubb:
Ping?

PR #39, opened 2015-05-26
C std headers

atombrella:
The function in 17e0154 is not compiled (the C++11 dictates that this signature is valid standard C++, but probably wasn't standard in 2003 when the comment was written.

I replaced a number of imports so they match the style in the C++ standard, e.g., <stdlib.h> => <cstdlib>, and since my editor deletes trailing whitespace, affected files, I also cleaned this up. 

include/boost/pending/fibonacci_heap.hpp has this snippet, but I'm uncertain whether <math.h> can also be changed into <cmath>

    #if defined(__sgi) && !defined(__GNUC__)
    # include <math.h>
    #else

### DIFF
--- a/example/astar-cities.cpp
+++ b/example/astar-cities.cpp
@@ -21,7 +21,7 @@
 #include <list>
 #include <iostream>
 #include <fstream>
-#include <math.h>    // for sqrt
+#include <cmath>    // for sqrt
 
 using namespace boost;
 using namespace std;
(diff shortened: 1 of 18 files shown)


atombrella:
@Belcourt could you please have a look at this?

atombrella:
@ericniebler because of the upcoming 1.59, can somebody please have a look at all these pull requests?

ericniebler:
Send a message to the Boost list and glad the Community Maintenance Team. I am not the maintainer of the BGL.

PR #40, opened 2015-06-04
Fix #11374 and #8791: issues with bundled properties and `find_flow_cost`

maelvalais:
I talked about that problem on [stackoverflow](http://stackoverflow.com/questions/30621255/using-named-parameters-and-bundled-properties-with-edmonds-karp-max-flow). This PR solves two different tickets because the cause was almost the same (bundled properties had not been tested thoroughly).  
Here are the tickets: 
- Issue [#11374](https://svn.boost.org/trac/boost/ticket/11374): `find_flow_cost()` won't work with bundled properties
- Issue [#8791](https://svn.boost.org/trac/boost/ticket/8791): no "flow" function (e.g. `edmonds_arp_max_flow()`) won't work with named parameters

I apologize for having mixed those two issues in one PR ; they were really "close" to each other... And I should have opened a new issue for the #8791 but it turned out that I didn't.

### DIFF
--- a/include/boost/graph/find_flow_cost.hpp
+++ b/include/boost/graph/find_flow_cost.hpp
@@ -14,9 +14,9 @@
 namespace boost {
 
 template<class Graph, class Capacity, class ResidualCapacity, class Weight>
-typename property_traits<typename property_map < Graph, edge_capacity_t >::type>::value_type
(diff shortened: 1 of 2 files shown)


maelvalais:
I'll create two different tickets because this PR won't be accepted

PR #43, opened 2015-07-21
Make it clearer that when tree_edge is called

xuhdev:
tree_edge function is called before the target vertex is discovered, so
users should be acknoledged that at this moment, the target's color is
still white.


Belcourt:
I don't find that wording to be clarifying so I'm closing this PR.  I understand your concern and will consider improving the documentation.

PR #45, opened 2015-10-07
[subgraph.hpp] add_vertex(u_global, g) on a subgraph does not recursi…

j-4:
…vely add to parent subgraphs

Hello,
One of the invariants of an induced subgraph is:
- If vertex u is in subgraph g, then u must be in g.parent().

This is true if you call add_vertex(g) on a subgraph to create a new vertex.
If you however choose to add a already existing global vertex to the subgraph, this addition is not propagated to the parents of this subgraph.

 I assume this is a bug and fixed it. 
Bug Ticket can be found here: https://svn.boost.org/trac/boost/ticket/11714
Thanks, Stefan

### DIFF
--- a/include/boost/graph/subgraph.hpp
+++ b/include/boost/graph/subgraph.hpp
@@ -366,6 +366,35 @@ struct graph_bundle_type<subgraph<Graph> >
 //===========================================================================
 // Functions special to the Subgraph Class
 
+
+namespace detail {
+    
+    template <typename G>
+    typename subgraph<G>::vertex_descriptor
+    add_vertex_recur_up(typename subgraph<G>::vertex_descriptor u_global,
+                subgraph<G>& g)
+    {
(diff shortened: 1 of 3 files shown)


murraycu:
If I were the maintainer (I'm not), I would prefer these commits to be squashed for easier reviewing.

murraycu:
https://svn.boost.org/trac/boost/ticket/11714 has a unit test for this. Shouldn't that be part of this pull request too?

Belcourt:
Any chance you could squash these and add the referenced unit test case?

j-4:
Hi,
I added the unit test into the testing framework and squashed the commits.
However, I have no idea to change the commit history here, so I generated a new pull request:
https://github.com/boostorg/graph/pull/80

Thanks for merging the bugfix :)
Let me know, if you need anything else!

PR #48, opened 2015-10-23
avoid requirement of boost::closed_plus in dijkstra_shortest_paths

ahhz:
dijkstra_shortest_paths requires closed_plus because of the details of
the relax function. By using a new relax_target function it no longer tries
to create distances above infinity and normal std::plus can be used.


murraycu:
Isn't that boost::closed_plus itself part of boost graph? What's the advantage to avoiding it?

ahhz:
Yes, boost::closed_plus is an undocumented part of boost graph.

Subjective reasons:
1. Getting to understand why BGL uses this closed_plus rather than simple std::plus is an unnecessary complexity. 
2. The recommended fix provides conceptual clarity: rather than doing something special when an INF and normal value are added, it avoids the situation altogether.
3. The closed_plus functor is ugly, because it only avoids overflow in one particular scenario and not others
4. The closed_plus functor is intertwined with the INF value, that are both input to the dijkstra functions. Changing one without adjusting the other accordingly will lead to errors.
5. Customisability: if you want to provide your own plus function, it needs to do the equivalent of std::plus, rather than the more complex boost::closed_plus

Objective reason:
1. Efficiency: some if-statements are elided every time two values are added.

murraycu:
Sounds good to me (I'm not the maintainer). I wonder if boost::closed_plus can be removed.

Belcourt:
Your relax_target seems to have omitted the case for undirected graphs where you sum the weight into the target vertex.  Why is that, or am I missing something?

murraycu:
That sounds like something that should have a test.

ahhz:
@Belcourt: I don't think that relax_target has omitted anything. 

The original relax function does two things: 

d(target) = min(d(target), d(source) + w(source, target);
if undirected:  d(source) = min(d(source), d(target) + w(target, source);

The relax target function only does the following:

d(target) = min(d(target), d(source) + w(source, target);

This is exactly what it is supposed to do, and avoiding the second part is the whole purpose of the relax_target function.

Note that when you pass an edge_descriptor to a function it contains the information about which vertex is the source and which is the target (even for undirected graphs where edges do not really have a source and a target: the context of how an edge is used does have a source and target).

See also the definition of the out_edges function on : 
http://www.boost.org/doc/libs/1_62_0/libs/graph/doc/IncidenceGraph.html

Belcourt:
I'll have to write a test case but I'm suspicious that you'll update the wrong vertex distance in the undirected case.

ahhz:
Do you know why you would think that? To me it is clear that the edge_descriptors that you get out of the out_edges function are 'correct' in the sense that the source(e) is the vertex of which we are considering all the adjacent vertices and that target(e) is always one of the adjacent vertices. In the dijkstra algorithm we only update the target of the edge, and hence any part of the relax function that updates the source can be removed from the relax_target function, which is what is being proposed here.

If your test would fail, I would think that your graph class is not a proper model of IncidenceGraph.

Belcourt:
I'm pondering this simple test case which shows that relax and relax_target produce different results.

For example: an undirected graph with d_u = 5, d_v = 4, and w_e = 0 results in:

relax updates d_u with 4, updates predecessor map, and returns true
relax_target doesn't update anything and returns false.

I don't doubt your assessment, I simply don't understand why this wouldn't break someone's code.  That's why I want to create the above as a test case and run it with both the old and new code.

ahhz:
Ah, I am not proposing a blanket replacement of relax for relax_target. However in the dijkstra algorithm dv is always greater then or equal to du. So, in that algorithm you can make the replacement, with the mentioned advantages.

anadon:
I'm helping out with the PR backlog. Looks like you have a code change.  Reading from the comments here, the changes may not be fully ready.  I would like some feedback on this proposed change.  This is to let you know and help me prioritize PR's.

ahhz:
As far as I can see, the proposed change is fully ready, but untested outside of my own project. The only remaining change to make is the [documentation](https://www.boost.org/doc/libs/1_68_0/libs/graph/doc/dijkstra_shortest_paths.html);  it should say that the default `CombineFunction` is `std::plus<D>` rather than `boost::closed_plus<D>`. 

You could make the equivalent changes for `dijkstra_shortest_paths_no_color_map`.

anadon:
Then please update the documentation and apply similar changes to `diijkstra_shortest)paths_no_color_map`.

ahhz:
OK, added those updates. I don't know what the issue with subgraph.html is. That was another fix that was somehow merged into this one.

anadon:
Thanks!  If you see other related cleanups or changes either let me know to add them to this or another PR as appropriate.  I'd like to get this repo fixed up.

anadon:
@ahhz Can you reabse on devel?

ahhz:
> 
> 
> @ahhz Can you reabse on devel?

I think I just did, but don't have much experience with git. Note that the "fixed type" and "coding style" fixes are not mine and were merged into my PR for some reason. I don't understand why because they are unrelated.

ahhz:
Ok, I will have another go. But could you be more precise in what you need me to do. What I did was

git rebase origin/develop alex_relax_target

And then resolve a trivial merge conflict through the github web-page.

Now I need to undo this and use a better command. Any suggestion on how I do that?

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

include/boost/graph/dijkstra_shortest_paths_no_color_map.hpp
```diff
@@ -97,7 +97,7 @@ namespace boost {
           !distance_compare(neighbor_vertex_distance, distance_infinity);
 
         // Attempt to relax the edge
-        bool was_edge_relaxed = relax(current_edge, graph, weight_map,
+        bool was_edge_relaxed = relax_target(current_edge, graph, weight_map,
```
anadon: Why change the function called?

anadon:
You're making this commit to reduce a header dependency, but don't remove any include.  Also, std::plus is C++14 and would bump the version requirement of graph.  There are also a large number of commits which look like they're formatting changes not necessarily related to this PR.  These are partly why it is harder for you to get things to work.

Let's get this logically sorted out.  What are you trying to achieve with that PR?  How is what is already here insufficient?  Why make these changes?

It looks like you're newer to coding and repo management so let's take this slow.

ahhz:
The original function was doing more than it should do. It checks whether to reduce the distance of the source  of the relaxed edge and that of the target of the relaxed edge, but in the dijkstra algorithm only the distance of the target edge can be reduced

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

anadon:
Sorry, misread the documentation.  It is older.

I see the 'avoid requirement of boost::closed_plus' but that is unrelated to what you are saying is incorrect behavior in dijkstra_shortest_path() from relax().  So please answer the contextual questions listed above.

anadon:
The use of closed_plus was to cope with weights of positive infinity and does not pull in an extra header.  This should not be changed.

relax() looks like it is doing what it should be doing.  How is relax() currently breaking your implementation?

ahhz:
It is not incorrect behavior but superfluous behavior. The consequence of that is a sum of target_weight plus edge_weight which necessitated using closed_plus. Getting rid of the superfluous stuff also gets rid of the closed_plus requirement. I think I responded to all concerns and questions. The problem is that those concerns were raised without consideration or understanding of how the dijkstra sp algorithm works.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

anadon:
If this is in the name of performance, then this PR will be declined.  That sort of superfluous behavior that some code needs is generally removed by the compiler.  Trust the compiler to optimize obvious things, and even some not so obvious things.

ahhz:
I have elaborately explained why I proposed this change. Did you not review the PR thread? Can we only propose improvements for code that is breaking implementations? Closed_plus is not to deal with edges of infinite weight, at least not in the dijkstra algorithm. If you read up on the documentation you will see that a requirement/assumption is that edge_weight < INF. Rather, in the dijkstra algorithm closed_plus dealt with vertices at infinite distance (which is not necessary when cutting superfluous branches from the relax function, which is what relax_target does).

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

ahhz:
Did you not read the PR thread? I have explained why I proposed this change as response to an earlier question.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

anadon:
Your PR does not fix incorrect behavior, and the possible performance improvements are already done by the compiler.

There is incorrect behavior introduced in the case that edge weight types roll over like an int, rather than hit a ceiling case like float.  This actually exposes a minor bug in closed_plus{} that I'll need to make a patch for.  On large graphs or graphs with large weights, the standard behavior desired differs from that of std::plus() because edges need to hit and stay at that maximum value.

Be less mean when challenged.  Insulting me does no good to anyone, and can poison a community.

ahhz:
How did I insult you?

anadon:
>The problem is that those concerns were raised without consideration or understanding of how the dijkstra sp algorithm works.

>Did you not read the PR thread? I have explained why I proposed this change as response to an earlier question.

Insinuating that I don't know basic algorithms and the accusatory questions.  If you are not aware how this came off, you really need to become aware.

But in any case, there is incorrect behavior which could be introduced so I'm disinclined to have this merged.

ahhz:
> Insinuating that I don't know basic algorithms

I would be highly surprised if you did not know the dijkstra shortest path algorithm. My point was that you and earlier commenters on this PR responded in a very literal way to the code, saying that my proposed change was not doing the same as the original code, i.e. relax != relax_target. My response has been that it is indeed not the same, but it is the same for the dijkstra algorithm, considering the way it works through the vertices in a particular order. If you go back in the PR thread you will see that that then stopped the discussion, which to me indicated an unwillingness to engage with the actual algorithm and a desire to just look at code equivalency. 

> the accusatory questions.

Yes, I see this comes across as unfriendly, apologies. At the same time, you must see that I can be a bit irritated when first I am asked to revisit and update a three year old PR (and it is actually older as I had something similar in Trac before, and then after doing so I get the same questions again that I answered in detail long before, and which has been recorded on this very same page.

ahhz:
> There is incorrect behavior introduced in the case that edge weight types roll over like an int, rather than hit a ceiling case like float

If the weight type is insufficient to hold the maximum distance, you have bigger problems and it would violate requirements / assumptions of the algorithm. For instance, it would fail to reach vertices that are in fact reachable.  

I don't know what bug you saw in closed_plus, but it  should be clear that this form of addition does not give general protection against overflows, it only treats the INF value as a special case, e.g. 0.7 * inf + 0.7 * inf would still overflow. 

I don't think this is a bug though, trying to snap all overflows in closed_plus to  INF (other than those involving the special INF value)  would be hiding problems instead of solving them.

What has it to do with rolling over? Aren't overflows equally problematic for int and float, except that for float you have additional problems due to limited precision?

anadon:
@ahhz Please rebase your PR on the upstream devel branch.

ahhz:
I tried too many thing trying to revert my incorrect changes, then incorrect reverts, etc. I have re-cloned the repository and made a fresh PR based on the current develop, it is #165 

This PR is therefore closed

PR #56, opened 2016-02-13
Remove trailing whitespace

atombrella:
Doesn't add any new functionality. There were tons of this in some files (`include/boost/graph/mcgregor_common_subgraphs.hpp` as one of the worst).


Belcourt:
This is an enormous commit that touches just whitespace.  Frankly, I'm okay living with a little extra whitespace so I'm not going to apply this PR.   I scanned the mentioned file but didn't find it objectionable though I did see a few lines that had 5 or 6 eol whitespace characters.

PR #61, opened 2016-03-02
Fix #11374 and #12038: issues with `find_flow_cost()`, bundled properties and named parameters

maelvalais:
I talked about that problem on [stackoverflow](http://stackoverflow.com/questions/30621255/using-named-parameters-and-bundled-properties-with-edmonds-karp-max-flow). This PR solves two different tickets because the cause was almost the same (bundled properties had not been tested thoroughly).  
Here are the tickets: 
- Issue [#11374](https://svn.boost.org/trac/boost/ticket/11374): `find_flow_cost()` won't work with bundled properties in conjonction to named parameters
- Issue [#12038](https://svn.boost.org/trac/boost/ticket/12038): every max-flow function (e.g. `edmonds_karp_max_flow()`) won't work with named parameters 

I apologize for having mixed those two issues in one PR ; I did try to separate them, but the last one (bundled properties and named parameters with `find_flow_cost()`) needed the two other commits.

### DIFF
--- a/include/boost/graph/find_flow_cost.hpp
+++ b/include/boost/graph/find_flow_cost.hpp
@@ -14,9 +14,9 @@
 namespace boost {
 
 template<class Graph, class Capacity, class ResidualCapacity, class Weight>
-typename property_traits<typename property_map < Graph, edge_capacity_t >::type>::value_type
+typename property_traits<Weight>::value_type
 find_flow_cost(const Graph & g, Capacity capacity, ResidualCapacity residual_capacity, Weight weight) {
-    typedef typename property_traits<typename property_map<Graph, edge_weight_t>::const_type>::value_type Cost;
+    typedef typename property_traits<Weight>::value_type Cost;
 
     Cost cost = 0;
     BGL_FORALL_EDGES_T(e, g, Graph) {
--- a/include/boost/graph/find_flow_cost.hpp
+++ b/include/boost/graph/find_flow_cost.hpp
@@ -28,7 +28,7 @@ find_flow_cost(const Graph & g, Capacity capacity, ResidualCapacity residual_cap
 }
 
 template <class Graph, class P, class T, class R> 
-typename property_traits<typename property_map < Graph, edge_capacity_t >::type>::value_type
+typename detail::edge_weight_value<Graph, P, T, R>::type
 find_flow_cost(const Graph & g,
                const bgl_named_params<P, T, R>& params) {
     return find_flow_cost(g,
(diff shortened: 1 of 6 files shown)


maelvalais:
By the way, do we need to write unit tests on each of these changes? I am a bit lost when it comes to UT on Boost; there doesn't seem to be any continuous integration or any script that builds and run all tests.

Belcourt:
Yes, there's ways to run all the tests.  Checkout boost, initialize the submodules, b2 headers, cd status, ../b2 -j8  or similar.  To run just graph tests, edit status/Jamfile.v2 and remove all the libraries except graph/test.

I would like to update the tests as all graph tests pass at present with these issues.  Can you add a bundled property test for both max-flow and find-flow-cost that fails with develop, and passes with your patch?

maelvalais:
Hi,
Thanks for reading my PR and for answering it! 
Ok, I'll try to write a test that would "reveal" the issue when tested on `develop`. The only slight issue I have is to find where I should put those tests in: 
- in `graph/test/max_flow_test.cpp`, meaning that I will add a new function that will use bundled properties (and calling `find_flow_cost()`) and another function that will use named parameters on `edmonds_karp_max_flow()`, one on `push_relabel_max_flow()` and one on `boykov_kolmogorov_max_flow.hpp`
- or should I create a specific file for each case, meaning that we would get 
  - `test_max_flow_with_bundled_params.cpp`
  - `test_find_flow_cost_with_named_and_bundled_params.cpp`

I'm a bit lost :/

Belcourt:
I think new test files would be fine.  Let me know if I can help.

maelvalais:
I just committed the first test on `find_flow_cost()` not working with bundled properties. `./b2 status` to launch the test. This is a compile-time test. 

As I rely on another code to generate the graph, so I had to modify it and make it more generic.

maelvalais:
I'm wondering: should I forget about testing the results (flow_cost=29) and only focus on revealing the compile-time flaw of `find_flow_cost`? 

Because I'm kind of re-doing the tests of `successive_shortest_path_nonnegative_weights_test.cpp`...

maelvalais:
I'm done with the two unit tests! I tested them before/after the modifications to be sure.

What do you think?

Belcourt:
This looks really good, thanks.  I'm testing them now and will push if clean.

maelvalais:
Btw, I thought it would be a good idea to have a continuous integration service as travis-ci or circle to run the tests automatically and see the results on the PRs. Here is my PR for that: #63

Belcourt:
So my only complaint with this patch is you use tabs instead of spaces.  Can you please indent with two spaces for each nested scope and don't use tabs?  Otherwise this is fine.

maelvalais:
Sure, sorry for that! 

Didn't read thoroughly the [coding policy](http://www.boost.org/community/policy.html) which specifically says not to use tabs 😕

maelvalais:
Oh, a conflict with `min_cost_max_flow_utils.hpp`... I'll take a look

maelvalais:
I (carefully) fixed the conflict and rebased. I could also "squash" the last fixup commit, as you wish.

PR #62, opened 2016-04-23
fix typo in document of subgraph

e-kwsm:


e-kwsm:
Not yet reflected.
http://www.boost.org/doc/libs/1_61_0/libs/graph/doc/subgraph.html

murraycu:
The releases are made from master. There's lots of stuff in develop that isn't in master.

PR #68, opened 2016-05-04
Add a .clang-format file.

murraycu:
So we can at least make the examples use a more consistent code style.

Run clang-format like so:
$ clang-format -i *.cpp
and use git diff to see what has changed.

This will strip trailing whitespace too. You can ignore that by
using git diff -w

---

This change is [Reviewable](https://reviewable.io/reviews/boostorg/graph/68)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,16 @@
+BasedOnStyle: WebKit
+Standard: Cpp11
+AlignAfterOpenBracket: false
+AlignEscapedNewlinesLeft: true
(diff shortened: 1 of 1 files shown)


murraycu:
This is just an initial suggestion. I don't see a very consistent coding style in Boost Graph, and I cannot find any general Boost C++ coding guidelines.

We should remove "SpacesInAngles: true" when BGL requires C++11.

maelvalais:
I just tested `clang-format`, I didn't know of that! Nice idea!

PR #69, opened 2016-05-05
Make files not executable.

murraycu:
See https://github.com/boostorg/admin/issues/47


murraycu:
These files are also executable in master, but not in develop, so hopefully they will become not executable when develop is merged into master:

./example/vf2_sub_graph_iso_example.cpp
./example/r_c_shortest_paths_example.cpp
./example/vf2_sub_graph_iso_multi_example.cpp
./example/sloan_ordering.cpp

PR #72, opened 2016-06-28
merged everyting from develop into master

murraycu:


murraycu:
Closing because Kenneth Belcourt would rather cherry-pick the commits. I prefer that too.

PR #78, opened 2016-10-27
Add degree to reverse_graph

mdb256:
degree is required for BidirectionalGraphConcept.
Also adds the concept check to the reverse_graph unit test.

### DIFF
--- a/include/boost/graph/reverse_graph.hpp
+++ b/include/boost/graph/reverse_graph.hpp
@@ -322,6 +322,13 @@ target(const detail::reverse_graph_edge_descriptor<Edge>& e, const reverse_graph
     return source(e.underlying_descx, g.m_g);
 }
 
+template <class BidirectionalGraph, class GRef>
+inline typename graph_traits<BidirectionalGraph>::degree_size_type
+degree(const typename graph_traits<BidirectionalGraph>::vertex_descriptor u,
+          const reverse_graph<BidirectionalGraph,GRef>& g)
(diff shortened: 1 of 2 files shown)


ilovezfs:
As I mentioned on [#29](https://github.com/boostorg/graph/pull/29#issuecomment-257316324), I'm aware of the following cases where the bug that this PR fixes has caused trouble, but there are likely many more:
https://github.com/01org/hyperscan/issues/39
https://github.com/bcgsc/abyss/issues/130
https://git.skewed.de/count0/graph-tool/issues/347

ilovezfs:
This patch (at least as adapted here https://raw.githubusercontent.com/ilovezfs/formula-patches/207b1060aa08bdbc2ada0cd7769f0934b5b521ea/boost/boost162-add-degree-to-reverse_graph.patch for the tarball) does not seem to work for graph-tool's build failure.

I'm still seeing
```
/usr/local/include/boost/graph/graph_concepts.hpp:131:13: error: no matching function for call to 'degree'
        n = degree(v, g);
```

Log here:
https://gist.github.com/ilovezfs/4bf6e521624d4757ed236e9d7c8522a3

ilovezfs:
This PR is fine. The problem was that graph-tool has its own modified filtered_graph.hpp

tkoeppe:
Is compressed_sparse_row_graph similarly affected? It looks like it's missing a `degree` function?!

PR #80, opened 2016-11-02
[subgraph.hpp] add_vertex(u_global, g) on a subgraph does not recursi…

j-4:
…vely add to parent subgraphs

Hello,
One of the invariants of an induced subgraph is:
- If vertex u is in subgraph g, then u must be in g.parent().

This is true if you call add_vertex(g) on a subgraph to create a new vertex.
If you however choose to add a already existing global vertex to the subgraph, this addition is not propagated to the parents of this subgraph.

 I assume this is a bug and fixed it. 
Bug Ticket can be found here: https://svn.boost.org/trac/boost/ticket/11714
Thanks, Stefan


j-4:
Is anyone still maintaining this?

PR #81, opened 2016-11-02
Examples: Use modern C++

murraycu:
These commits change the code in examples/ to use modern C++, specifically C++11. For instance, to use auto to avoid the verbosity of the traits and typedefs.

Even if we don't want to require C++11 for the API itself, maybe it's OK to use it for the example code, to more clearly demonstrate the API. I would understand if you don't want this.

The last commit ("Add and use make_range_pair") is even more debatable: It uses a range_pair utility class to allow use of range-based for loops, but maybe this should actually be a change to the API, or maybe it's something that will be better one day with C++ ranges in C++20.


Belcourt:
I'm currently disinclined to apply this PR though I do appreciate your efforts Murray.

jeremy-murphy:
I would love to see Boost.Graph get a modernization, I would actually vote for C++14 at this point.
Since Boost.Graph is not depended on by other libraries in Boost but it has been around for a long time, what would it take for a modernization to be accepted?

jeking3:
@Belcourt why are (were) you disinclined?  You rejected the PR but gave no reason for the submitter to allow for more improvement.

anadon:
Should we actually make this a branch for devel-cpp17?

murraycu:
> I'm helping out with the PR backlog. Looks like you have a large code change. I would like to merge this in around v1.69. The modernization is greatly appreciated, but making some fixes available and then having this change at a later version. 
[snip]

That's encouraging. Thanks. I don't quite understand what you'd like to do and when. Please feel free to request changes or updates from me that would help you to get this merged into the real code base.

anadon:
@murraycu Can you please rebase this PR on upstream's devel branch?

anadon:
@murraycu Looks like there are some CI failures.  Could you please take a look at those?

anadon:
2 failures are in the code changes, 2 are from external parameter parsing.  Shouldn't be that bad to fix.

murraycu:
At the moment I can't even build the regular develop branch:

[murrayc@murrayc-ThinkPad-X220 boost (develop *)]$ ./b2 
Performing configuration checks

    - default address-model    : 64-bit
    - default architecture     : x86

Building the Boost C++ Libraries.


    - symlinks supported       : yes
error: at /home/murrayc/checkout/github/boost/tools/build/src/kernel/modules.jam:107
error: Cannot create link boost/pending/property_serialize.hpp to libs/graph/include/boost/pending/property_serialize.hpp.
error: Link previously defined to another file, libs/graph_parallel/include/boost/pending/property_serialize.hpp.

murraycu:
The remaining build errors in CI are in tests/isomorphism.cpp, and seem to be unrelated to my changes.

anadon:
I've been out a while, but I think an external dependency changes and all
of graph broke.  I'd lets things settle while graph gets fixed back up.

example/range_pair.hpp
```diff
@@ -0,0 +1,48 @@
...
+// http://www.boost.org/LICENSE_1_0.txt)
+//=======================================================================
+
+#ifndef BOOST_GRAPH_EDGE_LIST_HPP
+#define BOOST_GRAPH_EDGE_LIST_HPP
+
+// Inspired by:
+// https://cplusplusmusings.wordpress.com/2013/04/14/range-based-for-loops-and-pairs-of-iterators/
+template <typename T_Pair>
+class range_pair
```
jeremy-murphy, line 16: I don't think this class is a good idea, sorry. For starters, it's not how the library is actually used. Secondly, there must be something equivalent in Boost.Range already, presumably `iterator_range`.

Thirdly, and more importantly, the best thing to do would be to add `begin(boost::pair)`/`end(boost::pair)` functions to `graph_traits.hpp` or some other common file so that `boost::pair` is automatically treated as a range. If that causes problems, an alternative would be to return `iterator_range` instead of `boost::pair`.

Now, that should be done separately from this PR so that this remains entirely about updating examples.

Finally -- the reason I am suddenly interested in this, is we have now effectively dropped support for pre-C++11, so we can actually consider merging!

murraycu:
>  we have now effectively dropped support for pre-C++11, so we can actually consider merging!

OK. Rebasing this is a bit awkwad after the reformatting in 2019, so I'll make a series of smaller pull requests instead.

Becheler:
I will close this PR as several smaller PRs have been merged ! Thanks again !

PR #52, opened 2015-11-26
Rank spanning branchings code for possible inclusion in Boost Graph Library

mbradle:
Hello,

I am submitting this implementation of the Camerini et al. algorithm to rank spanning branchings in directed graphs for possible inclusion in the Boost Graph Library.  The algorithm extends the Edmonds and Chu/Liu algorithm for finding the optimum branching to find the k optimum branchings (in order).  The complexity is O(kElog(V)).  I have included two example codes (example/rank-branchings1.cpp and example/rank-branchings2.cpp) that demonstrate usage.  I have also included two test codes (test/test_int_ranking.cpp and test/test_real_ranking.cpp) that construct (by brute force) all branchings in random complete graphs and then compare their ordering to the output from rank_spanning_branchings().  The test script test/test_spanning_branchings.sh executes these test codes and can be executed as

**./test_spanning_branchings 4**

where 4 would be the number of vertices in a random complete graph.  Finally, I have included a documentation file (doc/rank_spanning_branchings.html).

I know there was some interest in [including](http://sourceforge.net/p/edmonds-alg/mailman/message/24951212/) an implementation of Edmonds' algorithm into Boost.  I think my implementation would be a good possibility as well as an extension to find the subsequent best branchings.  In any event, thanks for your consideration, and best wishes.

Brad Meyer

PR #82, opened 2016-11-06
Add rank spanning branchings routine for possible inclusion in library.

mbradle:
Hello,

I am opening a new version of this pull request ([previous version](https://github.com/boostorg/graph/pull/52)).  I've squashed all the development commits into a single commit (but have included the development history in the commit message).

The documentation file (rank_spanning_branchings.html) describes the added code (rank_spanning_branchings.hpp).  I've included two examples (rank-branchings1.cpp and rank-branchings2.cpp), the second of which reads an input file (branching_input.txt).  I've also included a test routine (rank_branchings_test.cpp).  I added the examples and the test to the corresponding Jam files.  These have small conflicts with the development branch that I thought it might be better for the maintainers to resolve.

Thanks for your consideration, and best wishes.

Brad Meyer


anadon:
I'm helping out with the PR backlog.  Looks like you have feature additions, new documentation, no test, and new examples will be needed.  Please add tests for your added features.  This is to let you know and help me prioritize PR's.

anadon:
I'm in the process of dealing with some other build and test failures.  When I'm done with those, I'll have you rebase your changes on upstream and that may fix your problems.  Thank you for following up!

anadon:
Can you rebase your PR?

mbradle:
I've squashed and reworded commits.  I've rebased to boostorg:develop.  I'm still seeing a couple of travis-ci failures that I can't reproduce on my end.  Thanks, and best wishes.

anadon:
You test around line 820 has a template faux pas, and is causing timeout errors.  Better, but you still need to put in just a little more work.  You're close!

anadon:
I'll get to a proper review when I can.  This week got really busy for me, sorry!

anadon:
@mbradle It appears that you're still not rebased on boostorg/graph's devel branch.  Please do so.

anadon:
@mbradle Looking more closely now.  It looks like you've done your job and everything does appear clear.  I'll need to take more time to look through though.  Please be patient while I catch up!

mbradle:
Sorry, I rebased again before I saw your last message.  I hope that's ok.  Best wishes.

include/boost/graph/rank_spanning_branchings.hpp
```diff
@@ -0,0 +1,1159 @@
...
+      BOOST_CONCEPT_ASSERT(( ReadablePropertyMapConcept<IndexMap, Vertex> ));
+      BOOST_CONCEPT_ASSERT(( ReadablePropertyMapConcept<WeightMap, Edge> ));
+
+      unordered_set<Edge> best_branching, empty_set;
+
+      shared_ptr< std::pair<Edge, weight_t> > p;
+
+      Edge e;
+
+      heap::fibonacci_heap<BranchingEntry<Edge, WeightMap, Compare> > Q;
```
jeremy-murphy: In the same tradition of other Boost.Graph algorithms, I would recommend trying to allow the user to choose what type of property map and min heap to use, and provide these choices as defaults.
Some users will have special-case problems where different data structures are better, and in the future there might simply be a better property map, or min heap in general.

You'll have to experiment to find the best way to do it, but in general by making these data structures parameters to the algorithm so that users can pass them in by reference and an overload without those parameters that chooses the defaults. It will require some experimentation and might have to be all or nothing.

The other benefit of letting the client pass in data structures by reference is that the final result of computation is in them for the client to query afterwards if they like. (Think, for example, the open and closed sets in A*.)

include/boost/graph/rank_spanning_branchings.hpp
```diff
@@ -0,0 +1,1159 @@
...
+
+      unordered_set<Vertex> unvisited_vertex_set;
+
+      Vertex root_vertex;
+
+      BranchingGraph C;
+
+      vertex_idx_t n = num_vertices(g);
+
+      disjoint_sets<Rank, Pred> W( rank, pred1 ), S( rank, pred2 );
```
jeremy-murphy: I understand the motivation for single-letter variable names like `W`, `S`, `C`, etc, but in my experience, it's better to use a descriptive name. Not everyone has the algorithm paper on hand to look up what the letters mean. And it makes the variables awfully hard to text search for in a file, such as in this review.
`g` for graph, `u` and `v` for vertices, etc I think is fine... there is no perfect rule about it.

include/boost/graph/rank_spanning_branchings.hpp
```diff
@@ -0,0 +1,1159 @@
...
+      typedef typename property_traits<WeightMap>::value_type weight_t;
+
+      BOOST_CONCEPT_ASSERT(( VertexAndEdgeListGraphConcept<Graph> ));
+
+      BOOST_CONCEPT_ASSERT(( ReadablePropertyMapConcept<IndexMap, Vertex> ));
+      BOOST_CONCEPT_ASSERT(( ReadablePropertyMapConcept<WeightMap, Edge> ));
+
+      unordered_set<Edge> best_branching, empty_set;
+
+      shared_ptr< std::pair<Edge, weight_t> > p;
```
jeremy-murphy: Why `shared_ptr`? Is that really necessary?

jeremy-murphy: Like any programming construct they are not bad intrinsically, but they are often misused when a simpler construct would suffice.

include/boost/graph/rank_spanning_branchings.hpp
mbradle: Thanks for this.  I use two priority queues--one to keep track of branchings in the ranking routine and one to keep track of edge weights in the best and next routines.  The former is just a regular priority queue (PriorityQueue concept), but the latter needs to be mergeable (MergablePriorityQueue concept).  Neither one is "visible" to the user, and the former queue will have had all its entries popped off if all branchings have been found, but I do think it makes sense to allow the user to choose, with the Fibonacci heap as the default.  I'll experiment with this (I think there are some examples in Boost Graph), and with the desirability of distinguishing between the two queues.

include/boost/graph/rank_spanning_branchings.hpp
mbradle: Thanks for this.  I used pointers to check for uninitialized or not found objects, but I think maybe boost::optional is better.  For example,

```suggestion
      boost::optional< std::pair<Edge, weight_t> > p;
```

The NEXT routine could then return the appropriate pair, if found.  If not found, p will be null (I guess really boost::none), as in the current implementation.  I have a branch that does that successfully.  Are there any caveats to boost::optional?

include/boost/graph/rank_spanning_branchings.hpp
mbradle: Thanks, I agree with this suggestion.  Going through and making the changes will probably only reinforce your point.

include/boost/graph/rank_spanning_branchings.hpp
```diff
@@ -0,0 +1,1159 @@
...
+	    best_branching,
+	    empty_set,
+	    empty_set
+	  )
+	);
+      }
+      else
+      {
+	return;
+      }
```
jeremy-murphy: Some of the indentation here seems to have gone awry?

include/boost/graph/rank_spanning_branchings.hpp
```diff
@@ -0,0 +1,1159 @@
...
+                             IndexMap& v_id,
+                             WeightMap& w,
+                             Compare& comp,
+                             Rank rank,
+                             Pred pred1,
+                             Pred pred2,
+                             unordered_set<Edge>& include_edges,
+                             unordered_set<Edge>& exclude_edges
+                           )
+    {
```
jeremy-murphy: This function, `next_spanning_branching`, is very long. I recommend you break it up along the lines of major sequences, loops and/or conditions.

anadon:
I have a couple of big, no change PR's you might want to keep an eye on.

PR #83, opened 2016-11-29
fix comments about boost::detail::vector_matrix

e-kwsm:
`vector_matrix` does not seem non-copyable, and a non-copyable object should be passed by _reference_ or _move_ but not _value_.
Anyway, this comment is confusing.

### DIFF
--- a/include/boost/graph/exterior_property.hpp
+++ b/include/boost/graph/exterior_property.hpp
@@ -14,8 +14,7 @@
 namespace boost {
 namespace detail {
     // The vector matrix provides a little abstraction over vector
-    // types that makes matrices easier to work with. Note that it's
-    // non-copyable, meaning you should be passing it by value.
+    // types that makes matrices easier to work with.
     template <typename Value>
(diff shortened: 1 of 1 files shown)


e-kwsm:
Sorry, I mistook.

anadon:
I'm helping out with the PR backlog. Looks like you have a documentation change.  This will be in the second batch of merged PR's. This is to let you know and help me prioritize PR's.

anadon:
I'm disinclined on this PR.  I don't think the comment changes are improvements.  @jzmaddock Should this one be closed?

jzmaddock:
Actually I think the OP is correct: the existing comment is both self-contradictory and in conflict with the code, so I'm more inclined to <shrug> and apply the patch - it would of course be better to fix the comment so that it reflects the original intent, but as no one (including the original author probably after all this time!) knows what that is, this might be the next best thing?

anadon:
Well, better is better than nothing.  Merge it in then?

PR #84, opened 2016-12-22
Fix compiler error with release builds on VS2015

nre-ableton:
When using MSVC compiler optimization, using `param_not_found()` causes compiler error [C4172][1]: returning address of local variable or temporary

[1]: https://msdn.microsoft.com/en-us/library/d0ws2xs1.aspx

### DIFF
--- a/include/boost/graph/named_function_params.hpp
+++ b/include/boost/graph/named_function_params.hpp
@@ -228,6 +228,7 @@ BOOST_BGL_DECLARE_NAMED_PARAMS
   };
 
   struct param_not_found {};
+  static param_not_found g_param_not_found;
 
   template <typename Tag, typename Args>
   struct get_param_type: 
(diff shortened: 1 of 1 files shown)


jcageman:
I got the exact compile error fixed here using 1.65.1. Why is this not in 1.65.1?

nre-ableton:
@jcageman According to the [changelog](http://www.boost.org/users/history/version_1_65_1.html), 1.65.1 does not contain an updated version of the graph module. Hopefully it will be shipped in the next release.

jcageman:
Thanks for the quick reply. This is necessary to get boost compiling with VS2015/2017 so i would argue its a pretty important fix (maybe just on its own without other graph changes). Could you at least ensure it gets the right priority for next release? For now i will fix this in our local version.

nre-ableton:
@jcageman Sorry, I'm not affiliated with the Boost project, I just made the patch. 😉

jcageman:
Oke, thanks for the patch at least!

PR #89, opened 2017-05-10
Conditionally replace deprecated/removed C++98 std::bind1st by std::b…

DanielaE:
…ind, std::auto_ptr by std::unique_ptr, and std::random_shuffle by std::shuffle. Inline typedefs from C++98 function adapters.

Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/include/boost/graph/astar_search.hpp
+++ b/include/boost/graph/astar_search.hpp
@@ -46,11 +46,12 @@ namespace boost {
 
 
   template <class Graph, class CostType>
-  class astar_heuristic : public std::unary_function<
-    typename graph_traits<Graph>::vertex_descriptor, CostType>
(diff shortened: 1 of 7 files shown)


murraycu:
(I am not the maintainer.) There is lots of code in BGL that should be updated to C++11. I wouldn't want lots of ifdefs in the code. I wish boost would just choose to target newer compiler versions so it can move forward properly. C++11 is not too new.

DanielaE:
Yeah!
Unfortunately, as long as Boost.Graph supports pre-C++11 compilers, I have to deal with the fact that new versions of msvc (in C++17 mode) no longer support the long-deprecated C++98 stuff that is removed from the C++17 standard.

PR #93, opened 2017-07-05
Fix writes/reads outside allocated memory in resource constraints shortest paths

pmateusz:
Found an example where existing algorithm performs writes/reads outside allocated memory when using the default allocator. Implemented repo as a test included in the pull request. Valgrind reports 38 reads/writes outside allocated memory when running the test against the original version of the algorithm.

I did not get a segmentation fault on my machine, but the algorithm returned a list of edges that do not form a path (example: 1-5 5-2 2-3 4-0).

Investigation
If an existing label is found dominated it becomes immediately deallocated. However, pointers to that label still may exist in other partial paths. Further new allocations cause appearance of invalid paths. The assumption that a pareto optimal path consists only of non-dominated labels and therefore deallocating them is not correct (safe) if the dominance function performs weak inequality checks (as suggested by original author and presented in examples).

Fix
* Replaced a raw pointer to `r_c_shortest_paths_label` with `boost::shared_ptr`. The algorihtm finds shortest paths based on feasibility/dominance criteria specified by user, so there is a risk that pointers may form a cycle. Taking this into account the algorihtm resets pointers as the final step. I considered using weak pointers in the `r_c_shortest_paths_label`, but there are examples where such approach do not work. Provided it as a comment inside the algorithm code.
* Replaced `ks_smart_pointer` by `boost::shared_ptr`. The former class did not offer the smart_pointer functionality and is not necessary.
* Removed the `b_is_valid` checks. It is possible to come up with examples, where this approach will report false negatives (provided an example in comments). It is the same reason why weak_pointers will not work.
* Added a test case to prevent regressions.

Valgrind does not detect any leaks or possible leaks when running the test against the fixed algorithm.

### DIFF
--- a/include/boost/graph/r_c_shortest_paths.hpp
+++ b/include/boost/graph/r_c_shortest_paths.hpp
@@ -13,6 +13,8 @@
 #include <vector>
 #include <list>
 
+#include <boost/make_shared.hpp>
+#include <boost/enable_shared_from_this.hpp>
 #include <boost/graph/graph_traits.hpp>
 #include <boost/graph/iteration_macros.hpp>
 #include <boost/property_map/property_map.hpp>
@@ -21,12 +23,12 @@ namespace boost {
 
 // r_c_shortest_paths_label struct
 template<class Graph, class Resource_Container>
-struct r_c_shortest_paths_label
+struct r_c_shortest_paths_label : public boost::enable_shared_from_this<r_c_shortest_paths_label<Graph, Resource_Container> >
 {
   r_c_shortest_paths_label
   ( const unsigned long n, 
     const Resource_Container& rc = Resource_Container(), 
-    const r_c_shortest_paths_label* const pl = 0, 
+    const boost::shared_ptr<r_c_shortest_paths_label<Graph, Resource_Container> > pl = nullptr, 
     const typename graph_traits<Graph>::edge_descriptor& ed = 
       graph_traits<Graph>::edge_descriptor(), 
     const typename graph_traits<Graph>::vertex_descriptor& vd = 
@@ -37,8 +39,7 @@ struct r_c_shortest_paths_label
     pred_edge( ed ), 
     resident_vertex( vd ), 
     b_is_dominated( false ), 
-    b_is_processed( false ),
-    b_is_valid( true )
+    b_is_processed( false )
   {}
   r_c_shortest_paths_label& operator=( const r_c_shortest_paths_label& other )
   {
@@ -50,12 +51,11 @@ struct r_c_shortest_paths_label
   }
   const unsigned long num;
   Resource_Container cumulated_resource_consumption;
-  const r_c_shortest_paths_label* const p_pred_label;
+  const boost::shared_ptr<r_c_shortest_paths_label<Graph, Resource_Container> > p_pred_label;
   const typename graph_traits<Graph>::edge_descriptor pred_edge;
   const typename graph_traits<Graph>::vertex_descriptor resident_vertex;
   bool b_is_dominated;
   bool b_is_processed;
-  bool b_is_valid;
 }; // r_c_shortest_paths_label
 
 template<class Graph, class Resource_Container>
@@ -63,7 +63,6 @@ inline bool operator==
 ( const r_c_shortest_paths_label<Graph, Resource_Container>& l1, 
   const r_c_shortest_paths_label<Graph, Resource_Container>& l2 )
 {
-  assert (l1.b_is_valid && l2.b_is_valid);
   return 
     l1.cumulated_resource_consumption == l2.cumulated_resource_consumption;
 }
@@ -73,17 +72,15 @@ inline bool operator!=
 ( const r_c_shortest_paths_label<Graph, Resource_Container>& l1, 
   const r_c_shortest_paths_label<Graph, Resource_Container>& l2 )
 {
-  assert (l1.b_is_valid && l2.b_is_valid);
   return 
     !( l1 == l2 );
 }
 
 template<class Graph, class Resource_Container>
 inline bool operator<
-( const r_c_shortest_paths_label<Graph, Resource_Container>& l1, 
+( const r_c_shortest_paths_label<Graph, Resource_Container>& l1,
   const r_c_shortest_paths_label<Graph, Resource_Container>& l2 )
 {
-  assert (l1.b_is_valid && l2.b_is_valid);
   return 
     l1.cumulated_resource_consumption < l2.cumulated_resource_consumption;
 }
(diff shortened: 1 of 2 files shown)


jeremy-murphy:
Hi Mateusz. This sounds like a pretty serious bug and an important fix. I'm not an expert on this algorithm, but I've looked pretty closely through the code and it all looks good.
Sorry, what is the "reporo" that you refer to as a test?
Have you done any benchmarks just to check that the change hasn't inadvertently been deleterious to performance? (I personally wouldn't care about minor variation.)
I'm not the maintainer, btw, I just decided to start actively caring about a library that I use.  :)

test/r_c_shortest_paths_test.cpp
```diff
@@ -199,6 +199,89 @@ class dominance_spptw
...
+                            const spp_spptw_marked_res_cont& res_cont_2 ) const {
+        return res_cont_1.time <= res_cont_2.time
+                && res_cont_1.cost <= res_cont_2.cost
+                && std::includes( res_cont_1.marked.begin(),
+                                  res_cont_1.marked.end(),
+                                  res_cont_2.marked.begin(),
+                                  res_cont_2.marked.end() );
+    }
+};
+
```
jeremy-murphy: I think some brief comments on these new classes would help the reviewers understand what they do.

include/boost/graph/r_c_shortest_paths.hpp
```diff
@@ -21,12 +23,12 @@ namespace boost {
 
 // r_c_shortest_paths_label struct
 template<class Graph, class Resource_Container>
-struct r_c_shortest_paths_label
+struct r_c_shortest_paths_label : public boost::enable_shared_from_this<r_c_shortest_paths_label<Graph, Resource_Container> >
```
jeremy-murphy: I've never used `enable_shared_from_this`; what does it do?

pmateusz: [enable_shared_from_this](https://stackoverflow.com/questions/712279/what-is-the-usefulness-of-enable-shared-from-this) is a C++11 construct. See the link, it is better explained than I could.

pmateusz:
Hi @jeremy-murphy, I answered to your questions below.
* By _reporo_ I mean repro case. Sorry for my typo. I fixed the comment. I was referring to the new test case I added.
* Good questions about performance tests. I didn't do them. Changes proposed are to replace the proprietary smart pointer with the boost one. Moreover, I am not proposing an improvement, rather a fix for critical issue, so possible performance impact should be justifiable.

jeremy-murphy:
You're right, and I didn't mean to cast any doubt about performance, it's just a standard question. I assume you would have noticed if it was 10x slower, which is the kind of accidental performance bug I mean.
Well, I hope a maintainer has time to review it soon, as I think it really deserves to be merged asap.
@Belcourt are you currently the right person to gently remind about these PRs?

pmateusz:
@jeremy-murphy, it seems there is a ticket https://svn.boost.org/trac10/ticket/11723 with a similar issue. I asked the owner for a suggestion about next steps.

anadon:
I'm helping out with the PR backlog. Looks like you have a bugfix, test, and no new examples will be needed.  This will be in the first batch of merged PR's.  I will need to take time to understand this issue and fix more.  This is to let you know and help me prioritize PR's.

Belcourt:
I'm having trouble building graph / develop following all these PR's.  Do you, by chance, test each PR by building and running graph, graph parallel, and MPI tests before pushing it?  It might be wise to slow down so we don't irreversibly break something.

anadon:
Right now I'm letting people know that something is happening, adding some documentation, and prioritizing PR's to mess with.  Some of these old PR's need to be rebased so I'm stirring things up a little bit to get the submitters to check in and to know things are happening.  I'll be reading, testing, merging, and testing again PR's I say will be in the first batch of merges

include/boost/graph/r_c_shortest_paths.hpp
```diff
@@ -21,12 +23,12 @@ namespace boost {
...
 {
   r_c_shortest_paths_label
   ( const unsigned long n, 
     const Resource_Container& rc = Resource_Container(), 
-    const r_c_shortest_paths_label* const pl = 0, 
+    const boost::shared_ptr<r_c_shortest_paths_label<Graph, Resource_Container> > pl = nullptr,
```
pdimov: nullptr is C++11, the C++03 tests will fail.

test/r_c_shortest_paths_test.cpp
```diff
@@ -199,6 +199,89 @@ class dominance_spptw
...
+struct spp_spptw_marked_res_cont {
+    spp_spptw_marked_res_cont(SPPRC_Example_Graph::vertex_descriptor v, int c = 0, int t = 0 ) : cost( c ), time( t ), marked() {
+        marked.insert(v);
+    }
+    spp_spptw_marked_res_cont& operator=( const spp_spptw_marked_res_cont& other )
+    {
+        if( this == &other )
+            return *this;
+        this->~spp_spptw_marked_res_cont();
+        new( this ) spp_spptw_marked_res_cont( other );
```
pdimov: Invoking the destructor and then the constructor is not a good way to write an assignment operator. Perhaps it works here, I haven't checked, but it's not a good practice and it would be better to rewrite it in a more conventional way.

include/boost/graph/r_c_shortest_paths.hpp
anadon: This was fixed in #114.

pdimov: It's better to fix it here, if possible. It's easier to review and merge PRs one by one (and it would be easier to revert them later if some prove problematic.)

anadon: You can.  I'm new and don't have write access so having one place to work on keeps things a tad more sane.  It fixing commit was 6808062c07be7e394223819ea703ccb2c98e84a1 .

pmateusz: Thanks, @pdimov, I reapplied @anadon commit that removes usage of `nullptr` and removes spaces from end of lines.

test/r_c_shortest_paths_test.cpp
pmateusz: @pdimov, I agree. I implemented proper assignment operators in tests whenever possible. Only one has left in https://github.com/boostorg/graph/blob/develop/include/boost/graph/r_c_shortest_paths.hpp#L47. See this is part of released Boost. It was done by the original developer of the feature to allow reassignment of fields marked as const in the struct definition.

I was not changing it deliberately, because it has been already released and it is not causing any bugs as far as I am concerned. It is quite unusual construct to use without a reason.

anadon:
Try rebasing on the current develop branch.  I think some of your changes were already added so it won't be trivial to merge.

pmateusz:
Recent TravisCI failure in a single build is a timeout in the mas_test.cpp test, which is unrelated to changes in this thread.

jzmaddock:
Are these changes in https://github.com/boostorg/graph/pull/114?  If so it can be closed now.

anadon:
These changes were included in #114 and so this should be closed.

PR #95, opened 2017-07-27
Replace /bin/bash with /bin/sh

asomers:
/bin/bash is a Linuxism.  /bin/sh is portable, and this script isn't
using any bash-specific features.

### DIFF
--- a/doc/grid_graph_export_svg.sh
+++ b/doc/grid_graph_export_svg.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 #=======================================================================
 # Copyright 2009 Trustees of Indiana University.


anadon:
I'm helping out with the PR backlog. BSD, Linux, Mac, and now Windows have bash support.  Why is this change needed?  This will be in the first batch of merged PR's. This is to let you know and help me prioritize PR's.

asomers:
Because while bash _can_ run almost everywhere, it isn't necessarily installed by default, and on many systems /bin/bash isn't the correct path.  /bin/sh is correct on every unix system.

anadon:
Could you provide an example?  I'd like a concrete need in order to immediately accept this.

asomers:
On the BSDs, bash is usually installed to /usr/local/bin/bash, and is not installed by default.  On OSX, it can vary based on whether you're using homebrew, macports, etc.  I've even heard of Linux distros that don't use /bin/bash, but I can't find any on Google right now.  Some programs use "#! /usr/bin/env bash", which is more portable.  But it's always better to use "#! /bin/sh" for scripts that are bourne-compatible, which this one is.

anadon:
Good enough for me.  I'll add this to the first batch of merges since it can fix things.

PR #96, opened 2017-07-31
Moved graph_test.hpp from include directory to test directory.

eldiener:


anadon:
This change looks like it could break things.  Could you please provide a more complete description of the change and tests to verify it does not break anything?

eldiener:
The graph_test.hpp is not used anywhere except in tests and is not documented as part of the graph interface.  There is no reason it should be part of the include directory, The boostdep tool to find dependencies normally searches the include directory of a library but not the test directory.

anadon:
Let me read into this to make sure everything is safe.  If so, I'll include this in the second batch of changes.

jzmaddock:
Looks sensible to me - closing and re-opening to trigger CI.

pdimov:
Did (and just merged) the same PR in #120 without seeing this one, sorry. :-)

pdimov:
(I needed to cut the dependency on Test because it interfered with the CMake config work.)

PR #99, opened 2017-11-20
Remove deprecated std::unary_function and std::binary_function from examples

pavelkryukov:
std::binary_function was removed in C++17 and is no longer available while compiling with MS Visual Studio 2017

### DIFF
--- a/example/astar_maze.cpp
+++ b/example/astar_maze.cpp
@@ -46,7 +46,9 @@ typedef boost::graph_traits<grid>::vertex_descriptor vertex_descriptor;
 typedef boost::graph_traits<grid>::vertices_size_type vertices_size_type;
 
 // A hash function for vertices.
-struct vertex_hash:std::unary_function<vertex_descriptor, std::size_t> {
+struct vertex_hash {
+  typedef vertex_descriptor argument_type;
+  typedef std::size_t result_type;
   std::size_t operator()(vertex_descriptor const& u) const {
     std::size_t seed = 0;
     boost::hash_combine(seed, u[0]);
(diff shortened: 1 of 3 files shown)


anadon:
I'm helping out with the PR backlog.  It may be depreciated but that doesn't mean that it is not of value.  Could you make a strong argument for the change?  This is to let you know and help me prioritize PR's.

pavelkryukov:
> It may be depreciated but that doesn't mean that it is not of value. 

It was _deprecated_ in C++11, now it is _removed_, meaning it is no longer valid C++17 code.

> is no longer available while compiling with MS Visual Studio 2017

It means the example won't compile with MS Visual Studio 2017 or any newer version.

anadon:
I have a lot of these to read on.  Could you please check for me that no where else in the code needs std::unary_function or std::binary_function?  If not, I think I'll include this is the first batch of merges.

pavelkryukov:
Excuse me, could you please confirm your affiliation with Boost org?

The only things `std::unary_function` and `std::binary_function` had in C++03 were type definitions for `argument_type`, `first_argument_type`, `second_argument_type` and `result_type`, therefore there is no change in behavior as I copied type definitions inside the classes.
Moreover, it is an example, therefore it is not included to any other file.

anadon:
Here is a link to the thread: https://lists.boost.org/Archives/boost/2018/08/243175.php
In sum, I'm getting together low hanging fruit and making consolidated pulls which someone trusted with write access will handle.

anadon:
Here is a link to the thread: https://lists.boost.org/Archives/boost/2018/08/243175.php
In sum, I'm getting together low hanging fruit and making consolidated pulls which someone trusted with write access will handle.

pavelkryukov:
Reopened pull request to trigger CI builds.

jzmaddock:
Yup, also uncontroversial IMO.  Merging.

PR #100, opened 2018-01-02
Most members of std::allocate are deprecated in C++17

DanielaE:
Replace them by their cousins from std::allocator_traits. Without that, heaps of deprecation warnings will fall onto humble users when compiling with MSVC 15 in C++17 mode.

Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/include/boost/graph/adjacency_matrix.hpp
+++ b/include/boost/graph/adjacency_matrix.hpp
@@ -499,7 +499,11 @@ namespace boost {
 #if defined(BOOST_NO_STD_ALLOCATOR)
     typedef std::vector<StoredEdge> Matrix;
 #else
+#if defined(BOOST_NO_CXX11_ALLOCATOR)
     typedef typename Allocator::template rebind<StoredEdge>::other Alloc;
+#else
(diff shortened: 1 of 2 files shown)


anadon:
I'm helping out with the PR backlog. Looks like you have a code change.  I don't like the preprocessor order and would rather have it use built in C++ standard features before Boost's.  Could you revise your PR to comply?  This is to let you know and help me prioritize PR's.

DanielaE:
It seems like you haven't fully grasped the PR. There are no Boost features involved at all - it's all about using C++ allocators from the C++11 era (where possible) rather than the obsolete stuff.

anadon:
Let me take another look here.  There is quite the back log I'm looking through.

anadon:
I'm going to have to dig into this more.  This will be handled after the first two batches of merges.

jzmaddock:
I haven't checked this through, but we do need a fix for C++17 - I'm going to close and re-open in the hopes of triggering a CI build.

jzmaddock:
Merged, should all be uncontroversial changes to bring things in line with C++17.

PR #108, opened 2018-05-15
fix type error in grid graph example

mywtfmp3:
Won't compile using `boost::array<int, 2>` as grid graph constructor arguments.

Change to `boost::array<std::size_t, 2>`.


anadon:
I'm helping out with the PR backlog. Looks like you have a bugfix, no test, and no new examples will be needed.  I would like you to add a new test that works with your new fix and does not work on previous versions.  This is to let you know and help me prioritize PR's.

anadon:
@jzmaddock Here's a quick doc merge.

PR #109, opened 2018-06-24
Hyperlink DOIs against preferred resolver

katrinleinweber:
Hello :-)

The DOI foundation recommends [this new resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). Yes, a bit ironic that they would change the URL of their service, but it's now [encrypted](https://www.ssllabs.com/ssltest/analyze.html?d=doi.org). Because the old links with the `dx` subdomain continue to work, there is no urgent need to do anything.

However, I'd hereby like to suggest to follow the new recommendation and update all static DOI links.

Cheers!

### DIFF
--- a/include/boost/graph/bron_kerbosch_all_cliques.hpp
+++ b/include/boost/graph/bron_kerbosch_all_cliques.hpp
@@ -84,7 +84,7 @@ namespace boost
 //          number = {1},
 //          year = {2006},
 //          pages = {28-42}
-//          ee = {http://dx.doi.org/10.1016/j.tcs.2006.06.015}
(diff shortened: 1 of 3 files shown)


anadon:
I'm helping out with the PR backlog. Looks like you have a documentation update. This will be in the second batch of merged PR's. This is to let you know and help me prioritize PR's.

jzmaddock:
@anadon : trivial PR's like this that don't change code can be pushed through right away - if you spot any more trivial ones like this can you ping me and I'll go ahead and merge?

katrinleinweber:
Cool, thanks!

PR #110, opened 2018-06-24
O(n^3) maximum weighted matching

yi-ji:
This pull request contains:

- `include/boost/graph/maximum_weighted_matching.hpp`
 An O(|V|^3) implementation of maximum weighted matching algorithm for general graphs, and a verifier using brute-force searching.

- `test/weighted_matching_test.cpp`
Tests on randomly generated graphs using the brute-force verifier.

- `example/weighted_matching_example.cpp`
Some simple usage examples.

The O(|V|^3) implementation refers to:
Eugene L. Lawler, _Combinatorial Optimization: Networks and Matroids_, Chapter 6.10
(Note that there are a few mistakes in the book.)

Thanks to @aaw for his good implementation of maximum cardinality matching that helped me clear my mind when starting.


anadon:
I'm helping out with the PR backlog. Looks like you have a feature, test, and a new example. This will be in the second batch of merged PR's. This is to let you know and help me prioritize PR's.

yi-ji:
@anadon Great news, really thank you for what you are doing!
Please let me know if you need anything :)

anadon:
@yi-ji Please rebase your PR on the current devel.  There are some CI tools which have been enabled and a number of important cleanups.

test/weighted_matching_test.cpp
```diff
@@ -108,6 +108,8 @@ void weighted_matching_test(typename graph_traits<Graph>::vertices_size_type num
                 ++num_edges;
         }
     }
+
+    //print_graph(g);
```
anadon: Should this be removed?

test/weighted_matching_test.cpp
```diff
@@ -147,8 +149,8 @@ int test_main(int, char*[])
     // you can also lower the max_num_v and/or max_num_e
     
     std::size_t max_num_v = 16;
-    std::size_t max_num_e = 20;
-    std::size_t batch_size = 100;
+    std::size_t max_num_e = 24;
+    std::size_t batch_size = 16;
```
anadon: How do you calculate the optimal test size?

include/boost/graph/maximum_weighted_matching.hpp
```diff
@@ -54,7 +54,7 @@ namespace boost
         {
             typedef boost::iterator_property_map<typename std::vector<T>::iterator, VertexIndexMap> type;
         };
-        typedef decltype(graph::detail::V_EVEN) vertex_state_t; // anonymous enum defined elsewhere
+        typedef unsigned short vertex_state_t;
```
anadon: Does this type ever change in graph?

test/weighted_matching_test.cpp
yi-ji: Yes, sorry didn't notice, will do

include/boost/graph/maximum_weighted_matching.hpp
yi-ji: It's defined as anonymous enum and used only in `graph/max_cardinality_matching.hpp`,
see: https://github.com/boostorg/graph/blob/develop/include/boost/graph/max_cardinality_matching.hpp#L29

anadon: Then could it be removed entirely and not just its use here?

test/weighted_matching_test.cpp
yi-ji: I didn't, because I am using a brute-force verifier, it can be time-consuming when edge num increases (exponential), and travis-ci has time limit

anadon: Why not optimize it locally first?

yi-ji: I can wait for as long as it takes on my machine, but I didn't know about travis-ci time limit

anadon: I'm sure we can increase the time limit if it increases correctness.  But you should be able to determine the rough upper and lower bounds of these variables for testing correctness.

include/boost/graph/maximum_weighted_matching.hpp
yi-ji: "Here" is `graph/maximum_weighted_matching.hpp`, this enum is defined (and used) in `graph/max_cardinality_matching.hpp`, not by me. 
Another idea is to assign a name for this enum in `graph/max_cardinality_matching.hpp` but I think maybe it's better not to tamper with existed files

anadon: I'm a little uncomfortable with changing something which seems to be related to a template parameter.  If this is just for testing time, it is probably best to leave it be.

test/weighted_matching_test.cpp
yi-ji: You're right, I will give a rough time estimate regarding edge num

include/boost/graph/maximum_weighted_matching.hpp
yi-ji: Did you mean leave `decltype(graph::detail::V_EVEN)` be? it won't compile by g++
I understand you concern, however `vertex_state_t` has only possible values of {0, 1, 2} (no chance to be changed by logic) and only used in `graph/max_cardinality_matching.hpp`, so unsigned short shouldn't cause any problem I think

test/weighted_matching_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include <boost/graph/maximum_weighted_matching.hpp>
+
+#include <iostream>                      // for std::cout
+#include <boost/property_map/vector_property_map.hpp>
+#include <boost/graph/adjacency_list.hpp>
+#include <boost/graph/adjacency_matrix.hpp>
+#include <boost/graph/random.hpp>
+#include <ctime>
+#include <boost/random.hpp>
+#include <boost/test/minimal.hpp>
```
jeremy-murphy: I strongly recommend that you don't use randomly generated test cases here. Using them as a separate utility to discover bugs is up to you, but the test cases here should attempt to provide good coverage of the algorithm by providing inputs that execute different code paths, edge cases, etc.
Then you also don't need to run a brute-force verification: just store the answer and compare.

yi-ji: Nice advice, thank you. I will remove brute-force verification and use pre-computed answers instead.

include/boost/graph/maximum_weighted_matching.hpp
anadon: OK.  This is new code so good enough for me at this point.  I'll read through the PR in more detail when the tests pass.

anadon:
It looks like the anonymous enum to short is causing some other knock off effects causing compilation failure.

yi-ji:
@anadon Not sure why jobs 87.17 and 87.18 failed because no output shown...seems related to clang++5.0? Though I tested locally without seeing anything wrong
Or might need help from this: https://docs.travis-ci.com/user/common-build-problems/#ruby-tests-frozen-and-cancelled-after-10-minute-log-silence

anadon:
@jzmaddock This looks good to me to merge.  It provides an interesting addition to BGL.

anadon:
@yi-ji Make sure to add documentation and examples.

jzmaddock:
We have an example, but we do need docs before merging.  Please add "[CI SKIP]" to pure documentation commit messages to stop CI builds cycling needlessly: we're short on CI testing time especially on appveyor which seems to all run through my account :(

anadon:
Would it help if I set up something through appveyor to run these through?

yi-ji:
@anadon @jzmaddock documentation added :)

anadon:
I'm not seeing CI run.  Do you have an idea of what is up with that?

yi-ji:
@anadon  I added "[CI SKIP]" into the commit message as @jzmaddock [asked](https://github.com/boostorg/graph/pull/110#issuecomment-435134621). No need to run a CI build for pure document commits, right?

yi-ji:
@jzmaddock Could you help merge this PR?

jeremy-murphy:
If you all don't mind, I would like to have a good look through the algorithm implementation before we merge. There is a lot of code there.
I see some use of `shared_ptr` and virtual functions that I'm a bit suspicious about.

And someone remind me... is Boost.Graph still on a strict C++03 diet?

jzmaddock:
>If you all don't mind, I would like to have a good look through the algorithm implementation before we merge. There is a lot of code there.
>I see some use of shared_ptr and virtual functions that I'm a bit suspicious about.

It would be great if someone would give this a once over.

>And someone remind me... is Boost.Graph still on a strict C++03 diet?

I'd prefer to keep existing code passing with existing compilers.  But for strict additions, I don't see why we can't use C++20 if that's the only way to do it.  As long as it's documented of course.  In any case this looks all C++03 to me at first glance, and the few things that have C++11 equivalents (shared_ptr, tie) are internal details so far as I can see.

doc/bibliography.html
```diff
@@ -438,6 +438,16 @@ <H2>Bibliography</H2>
...
+<p></p><dt><a name="edmonds65:_max_weighted_match">74</a>
+<dd>J. Edmonds<br>
+<em>Maximum Matching and a Polyhedron with 0, 1-Vertices</em><br>
+Journal of Research of the National Bureau of Standards B 69, pp. 125-130, 1965.
+
+<p></p><dt><a name="gabow90">75</a>
+<dd>Harold N. Gabow<br>
+<em>Data Structures for Weighted Matching and Nearest Common Ancestors with Linking</em><br>
+Proceedings of the First Annual ACM-SIAM Symposium on Discrete Algorithms, pp. 434-443, 1990.
+
```
jeremy-murphy: I know this publication from 1990 is more recent, but maybe his original paper on the implementation is more relevant and specific? https://dl.acm.org/citation.cfm?id=321942

doc/maximum_weighted_matching.html
```diff
@@ -0,0 +1,163 @@
...
+<a name="sec:weighted_matching">Before you continue, it is recommended to read
+about <a href="./maximum_matching.html">maximal cardinality matching</a> first. 
+A <i>maximum weighted matching</i> of an edge-weighted graph is a matching
+for which the sum of the weights of the edges is maximum. 
+Two different matchings (edges in the matching are colored blue) in the same graph are illustrated below. 
+The matching on the left is a maximum cardinality matching of size 8 and a maximal
+weighted matching of weight sum 30, meaning that is has maximum size over all matchings in the graph
+and its weight sum can't be increased by adding edges.
+The matching on the right is a maximum weighted matching of size 7 and weight sum 38, meaning that it has maximum
+weight sum over all matchings in the graph.
```
jeremy-murphy: In terms of nomenclature, Gabow (1990) mentions a _maximum weight matching_ problem but in his earlier 1976 paper the problem is called _maximum matching_, as it is by Edmonds in his original paper.
My guess is that he added "weight" in to differentiate against the _maximum cardinality matching_ problem.
But either way, I'm going to suggest that it is "weight", not "weighted", although I know he uses the latter in some other names. Anyway, it's not a serious issue, just my pedantic opinion.

In addition to that, the 1990 paper raises another point, that this O(n^3) algorithm is not the only algorithm that solves maximum weight matching, as it presents an O(n(m + n log n)) algorithm. It's perfectly reasonable to implement the O(n^3) algorithm, but we need to name it such that we can also name the newer algorithm, e.g.:

- `brute_force_maximum_weight_matching` -- O(?)
- `label_encoding_maximum_weight_matching` -- O(n^3)
- `blossom_expansion_maximum_weight_matching` -- O(n(m + n log n))

They are just ideas, I'm not sure if they are entirely accurate, especially the last one.

Then, maybe, we could have a `maximum_weight_matching` that uses some heuristics to pick the fastest algorithm for a given problem size. Just an idea, not important.

doc/bibliography.html
yi-ji: I agreed, change will be made. 
Just to be clear, the labeling method is not generalized to weighted matching in the paper you mentioned, though it does can be. 
I have to admit that the 1990 paper focused more about improving the time complexity from O(n^3) to O(n(m+nlogn)) using link-cut-trees to find nearest common ancestor, which is not implemented in this PR..

doc/maximum_weighted_matching.html
yi-ji: I believe both "maximum weight / weighted matching" are commonly used. In fact, another large C++ template library for network algos, [LEMON](http://lemon.cs.elte.hu/pub/doc/latest/a00254.html), uses `MaxWeightedMatching`.

Thanks for the name prefix suggestion. However the examples might not be quite accurate. The O(n^3) and O(n(m + n log n)) algorithms both have labeling and blossom expansion operations. The only difference is the O(n(m + n log n)) algorithms uses link-cut-tree to find nearest common ansestors, which is not important/radical/obvious enough to call them with different names.
I didn't implement the O(n(m + n log n)) algo mainly because there is no available link-cut-tree module in boost. If it becomes available and someone wants to improve to O(n(m + n log n)) someday, he/she could just make a few lines of code change and doesn't have to keep an O(n^3) version.

doc/bibliography.html
jeremy-murphy: I'm not quite sure what you mean about the labeling method is not generalized in that paper?

But otherwise, glad my suggestion was helpful.

I didn't realize that NCA (which I think is typically called LCA?) is part of this algorithm, although I haven't finished going through it yet. Where is it in here?

doc/maximum_weighted_matching.html
jeremy-murphy: Ah, hmmm, OK, you think the difference between the O(n^3) and O(n(m + n log n)) algorithms will be a matter of passing in different LCA sub-algorithms? That would certainly make the naming easier.

Funnily enough, I am actually working on... close to finishing... a pragmatic implementation of LCA, but it's not Gabow's  link-cut version.  When I have a chance, I'll have a look at how much more work it would be to extend it, although I want to finish the basic implementation first.

doc/bibliography.html
yi-ji: I mean, the labeling procedure in the paper you mentioned only applies to unweighted matching. The author just claimed at the end that it could be generalized to weighted matching.

Link-cut-tree based NCA/LCA is not implemented in this PR. I am using a trivial method, which is used in `max_cardinality_matching.hpp` as well.

In fact, this 1990 citation should not be deleted... I didn’t look carefully where it is cited when replying you. It’s not cited for algorithm reference, but for pointing out the best known time complexity is not O(n^3).
I cited Edmond: “Maximum Matching and a Polyhedron with 0, 1-Vertices” for my implementation. 
Of course I will still add the paper you mentioned, because Edmond’s paper did not elaborate implementation detail for labeling procedure.

doc/maximum_weighted_matching.html
yi-ji: Thanks. More funnily... I am also working on and close to finishing a link-cut-tree module😂 but I believe it’s okay if we open dIfferent PR. Let’s see what to do then.

yi-ji:
Review comments resolved. Ready for merge again! (If no more of them)

jeremy-murphy:
I haven't had a chance to look at the code itself in detail yet and realistically I won't until the new year. I'm inclined to say that so long as it is well tested (which I think it is), let's not let perfect be the enemy of good, merge it now, and look at refactoring when we have time.

yi-ji:
Happy 2019 guys.. so is there anything stalling the merge...?

jzmaddock:
Merged.
I've re-ordered the initialization of members of maximum_weighted_matching to avoid pages and pages of gcc warnings.  Also added an entry to the documentation table-of-contents.

PR #111, opened 2018-07-30
Null pointer when reindexing edge list

raahlb:
Don't copy value_type if not needed, as source's m_property will be set to null when copied.

See https://svn.boost.org/trac10/ticket/13544.


jeremy-murphy:
Would it be relatively easy to add a test-case that fails without this fix?

raahlb:
I added a test case. Also added it to the Jamfile.v2. Not used to this test system, so please verify I did it in the right way.

anadon:
I'm helping out with the PR backlog.  Looks like you have a bugfix, test, and no new examples will be needed.  This will be in the first batch of merged PR's.  This is to let you know and help me prioritize PR's.

anadon:
your test file was not included in the test directory.  Make sure you added it to git.

anadon:
I see what happened.  The test you added has a different file name than the test file you added.

raahlb:
Saw now that you had already fixed it yourself. Not super important, but the name "delete edge" is a bit misleading, it's actually a vertex that is being deleted in the test. Sorry about that.

Thanks!

jzmaddock:
Are these changes in https://github.com/boostorg/graph/pull/114?  If so this can be closed now.

anadon:
These changes are in #114 and so this should be closed.

PR #118, opened 2018-09-25
fix visibility of exceptions within Boost.Graph

apolukhin:
This is required to avoid issues with `-fvisibility=hidden` builds (those are now default in Boost develop branch).

More info on the macro available [here](https://www.boost.org/doc/libs/1_68_0/libs/config/doc/html/boost_config/boost_macro_reference.html#boost_config.boost_macro_reference.macros_for_libraries_with_separate_source_code.macros_controlling_shared_library_symbol_visibility)

Related issue: https://github.com/boostorg/graph_parallel/pull/16

### DIFF
--- a/include/boost/graph/bipartite.hpp
+++ b/include/boost/graph/bipartite.hpp
@@ -32,7 +32,7 @@ namespace boost {
      */
 
     template <typename Vertex>
-    struct bipartite_visitor_error: std::exception
+    struct BOOST_SYMBOL_VISIBLE bipartite_visitor_error: std::exception
     {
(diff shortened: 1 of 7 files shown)


anadon:
The relaxed heap failure is currently addressed in PR #114.  It might actually be related to why #114 is failing Mac builds, and if so I'll look into merging this in.  Thanks for helping out!

anadon:
It looks like this kind of change doesn't change code functionality, only linking.  It also appears that this kind of change could be applied more broadly.  Are there other things which should be guarded with `BOOST_SYMBOL_VISIBLE`?

PR #121, opened 2018-10-17
Fix the copy constructor of the subgraph.

kenkangxgwe:
In the previous version, the children and their properties are not copied correctly.

### DIFF
--- a/include/boost/graph/subgraph.hpp
+++ b/include/boost/graph/subgraph.hpp
@@ -131,12 +131,24 @@ public:
 
     // copy constructor
     subgraph(const subgraph& x)
-        : m_parent(x.m_parent), m_edge_counter(x.m_edge_counter)
-        , m_global_vertex(x.m_global_vertex), m_global_edge(x.m_global_edge)
+        : m_parent(x.m_parent), m_edge_counter(0)
     {
         if(x.is_root())
         {
-         m_graph = x.m_graph;
+            m_graph = x.m_graph;
+            m_edge_counter = x.m_edge_counter;
+            m_global_vertex = x.m_global_vertex;
+            m_global_edge = x.m_global_edge;
+        }
+        else
+        {
+            get_property(*this) = get_property(x);
+            typename subgraph<Graph>::vertex_iterator vi,vi_end;
+            boost::tie(vi, vi_end) = vertices(x);
+            for(; vi != vi_end; ++vi)
+            {
+                add_vertex(x.local_to_global(*vi), *this);
+            }
         }
         // Do a deep copy (recursive).
         // Only the root graph is copied, the subgraphs contain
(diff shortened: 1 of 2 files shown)


anadon:
We recently had a PR, #80, which touched on something close to this you might want to look at.  Before this gets accepted, please add a test case which illustrates the problem, doesn't work on the past incorrect behavior, and works with your PR.

kenkangxgwe:
I did not see the #80's relation with the copy ctor. Current copy ctor does not recursively copy the children and properties. I have added a test case in `test/subgraph.cpp`, you may check it.

anadon:
It was a quick thought.  I've been a little preoccupied.

include/boost/graph/subgraph.hpp
```diff
@@ -131,29 +131,35 @@ class subgraph {
...
     subgraph(const subgraph& x)
-        : m_parent(x.m_parent), m_edge_counter(x.m_edge_counter)
-        , m_global_vertex(x.m_global_vertex), m_global_edge(x.m_global_edge)
+        : m_parent(x.m_parent), m_edge_counter(0)
     {
         if(x.is_root())
         {
-         m_graph = x.m_graph;
+            m_graph = x.m_graph;
+            m_edge_counter = x.m_edge_counter;
```
anadon: If the subgraph is actually the entire graph keep count of its number of edges, else don't keep track of the member graph's number of edges?  This seems like a change in function.

test/subgraph.cpp
```diff
@@ -18,6 +18,27 @@
 
 // UNDER CONSTRUCTION
 
+
+// This is a helper function to recusively compare two subgraphs.
```
anadon: Please add more detail about what it is testing, and the motivation for testing it.

test/subgraph.cpp
```diff
@@ -110,6 +131,56 @@ int test_main(int, char*[])
         }
     }
 
+    // This block generates a complete graph with 8 vertices,
+    // and puts the first and last four of the vertices into two children.
+    // Do these again to the children, so there are 4 grandchildren with 2 vertices for each.
+    // Use the copy constructor to generate a copy and compare with the original one.
```
anadon: What about subgraph edges?  Those also need testing.

kenkangxgwe: I only checked the numbers. How do I compare two edge descriptors from two different graphs?

include/boost/graph/subgraph.hpp
kenkangxgwe: If the `m_graph` is a child of its parent, we manually add all the vertices from `x.m_graph` to `m_graph`. This will automatically increase the `m_edge_counter` instead of assign `x.m_edge_counter` to it.

test/subgraph.cpp
anadon: You could load the structurally identical graphs with each node holding a unique ID in that graph identical to the equivalent edge in the other graph.

kenkangxgwe: Ok, I just added a comparison of two corresponding edge indices

anadon:
I'm going to try to revisit this Wednesday.  I'm overloaded with work right now and can't give a review due diligence.

kenkangxgwe:
Take your time and have a nice weekend

anadon:
@jzmaddock This PR looks good to me.  It timed out on one test which could be rerun.

jzmaddock:
I've pushed a fix for the timeout to develop .travis.yml - I suspect the PR will just timeout again if we re-run, so I'll go ahead and merge and we'll see.

PR #125, opened 2018-10-23
WIP: Fixes #122

### LINKED ISSUE #122: Enhancement proposal: verices() and edges() returning boost::iterator_range
Currently the `vertices()` and `edges()` functions in BGL return an `std::pair` of iterators. To iterate over vertices, a code like following is usually written:
```c++
typename graph_traits<Graph>::vertices_size_type b = 0;
typename graph_traits<Graph>::vertex_iterator i, end;
for (boost::tie(i, end) = vertices(g); i != end; ++i)
    do_something(*i);
```

If `vertices()` and `edges()` returned a [boost::iterator_range](https://www.boost.org/doc/libs/1_55_0/libs/range/doc/html/range/reference/utilities/iterator_range.html), a cleaner code could be written:
```c++
for (auto v : vertices(g))
    do_something(v);
```

mkatliar:
I have changed the `vertices()`, `edges()`, `adjacent_vertices()`, `in_edges()`, `out_edges()` functions such that they return an object which inherits `std::pair<Iterator, Iterator>`, and additionally models a container. This allows the use of range-based for loops with the results of those functions, and keeps compatibility with the existing code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ kevin-bacon2.dat
 random.dot
 triangular-fr.dot
 triangular-kk.dot
+.vscode
--- a/include/boost/graph/adjacency_matrix.hpp
+++ b/include/boost/graph/adjacency_matrix.hpp
@@ -24,6 +24,7 @@
 #include <boost/mpl/bool.hpp>
 #include <boost/graph/adjacency_iterator.hpp>
 #include <boost/graph/detail/edge.hpp>
+#include <boost/graph/detail/graph_iterator_range.hpp>
 #include <boost/iterator/iterator_adaptor.hpp>
 #include <boost/iterator/filter_iterator.hpp>
 #include <boost/range/irange.hpp>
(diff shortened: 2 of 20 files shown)


jeremy-murphy:
It's great that you have maintained backward compatibility, but changing those interfaces is not something to be done lightly. This looks like a lot of code for marginal benefit because, as I mentioned in the issue tracker, one can simply use `boost::for_each` (or any other algorithm) from Boost.Range.
Also, it's adding what looks like a non-trivial amount of type computation to what was previously a very simple `std::pair`. I try not to care about compilation time in general, but I guess when it's something that I don't need, then I care.  :)

Another way to solve this problem would be to have a `boost::pair` equivalent data structure to `std::pair`, because then the simple solution would be to add overloads of `begin`/`end` functions to the boost namespace for `pair` of iterators:
```C++
namespace boost {
  template <typename T>
  T begin(pair<T, T> const &iterators) {
    return iterators.first;
  }

  template <typename T>
  T end(pair<T, T> const &iterators) {
    return iterators.second;
  }
}
```
You could also add a bunch of interface checking with `enable_if` to ensure that `T` is an iterator, etc.
There's a lot of work there to add a boost::pair, but the final result would be more elegant and maintain the low overhead that the current implementation has.

mkatliar:
@jeremy-murphy I like the solution with `boost::pair`, it is indeed much simpler. `boost::pair` is something that does not exist yes, does it?

jeremy-murphy:
I guess we could add a `pair` class in some private Boost.Graph namespace that simply inherits from `std::pair` and put those `begin`/`end` functions in that namespace.
One day there might be call for a general-purpose pair class in Boost but for now all we want is `std::pair` in a namespace that we control.
@jzmaddock does that sound reasonable?

jzmaddock:
I confess I'm somewhat wary of this as it's a breaking interface change: I admit there probably isn't much code it would break, but I'm sure it would break some.

While this is nice to have, using:

```
using boost::make_Iterator_range;

for ( auto edge: make_iterator_range(my_graph.edges()))
{
   // do something.
}
```

Isn't that much harder?

I do see the attraction though and changing pair to class_derived_from_pair isn't such a big change, I guess the question is whether a breaking change like this is something that the CMT should be doing?  Maybe raise it on the mailing list?

PR #131, opened 2018-11-06
Ear decomposition

anadon:
I'm going through other forks and came across this.  I think it is worth a look.


anadon:
I think I have to merge in develop, then update the pull.

jeremy-murphy:
@j-4 I know six years is a long time, but we can get this merged into Boost.Graph if you're motivated?

PR #132, opened 2018-11-06
Dev3

anadon:
I'm going through other forks and came across this.  I think it is worth a look.


anadon:
Didn't see #82 when looking at this,

PR #135, opened 2018-11-09
Appletest -- not a merge request

felix-salfelder:
#86 fails to compile on apple (libc++?). just making sure it is not the bugfix...


jeremy-murphy:
You still testing something?

felix-salfelder:
No, I must have given up with apple when CI took days to even start.

It was all about #86, which is still open...

PR #144, opened 2018-12-16
Fix up tests and examples so that more are tested

jzmaddock:
Also fix MSVC security errors, and document which examples are broken and need revising.


jeremy-murphy:
Wow, it's like a Trojan horse of API changes just exploded.

jzmaddock:
> Wow, it's like a Trojan horse of API changes just exploded.

Nod: the examples were never tested and when stuff got changed - years ago I suspect - the breakages were never noticed or fixed :(

anadon:
I'm still trying to watch all these.  I'm still off getting better with programming before I become active again.

PR #152, opened 2018-12-20
 Fix regressions & update .yml scripts

CromwellEnage:
<boost/graph/named_function_params.hpp>
* Original fix worked only when one boost::parameter::keyword was bound and passed in.  This fix now works when binding and passing in one or more boost::parameter::keyword objects.

<boost/graph/isomorphism.hpp>
* Make isomorphism_impl a metafunction that returns the function object result_type for the fix to work.

"test/isomorphism.cpp"
* Test the fix with two boost::parameter::keyword objects.

".travis.yml"
"appveyor.yml"
* Update as per Peter Dimov's message announcing the merging of CMake into boostorg/develop.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -274,6 +274,8 @@ install:
   - cd ..
   - git clone -b $BOOST_BRANCH --depth 1 https://github.com/boostorg/boost.git boost-root
   - cd boost-root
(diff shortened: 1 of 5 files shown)


jzmaddock:
Apologies, I missed this one, looks good, so merging.

PR #160, opened 2019-02-05
successive_shortest_path_nonnegative_weights const bugfix

wygos:
THX to Ireneusz Szcześniak for finding this bug


Becheler:
Thank you for opening this PR ! It was superseded by #508

PR #168, opened 2019-04-24
Fix for a number of broken links.

tinko92:
Hi,

sorry for this pretty large commit. It is intended to fix a large number of broken external links in the Boost.Graph library. Here is what it does:

1. Replaces all instances of http://www.sgi.com/tech with https://boost.org/sgi .
2. Replaces all instances of http://www.osl.iu.edu/~lums with https://homes.cs.washington.edu/~al75 . I hope, this is appropriate, it is his most current homepage I could find.
3. Replaces all instances of http://www.cs.rpi.edu/~beevek with http://cs.krisbeevers.com (See 2) )
4. Replaces links to the old qhull websites with links to qhull.org
5. Replace two broken deep links to pdfs of papers with links to the ResearchGate page holding the full-text pdfs.
6. Numerous links that are dead were prefixed with webarchive-links, so the old content is shown. I hope the chosen dates are appropriate but ultimately I was not able to confirm whether I've linked the correct versions so I used my best judgment.
7. Replaced a dead link to an e-mail in the boost mailing list archive with a link to (almost surely) the same e-mail in the current archive. The old e-mail is not in the web archive, but if the list of alternative urls in this https://www.mail-archive.com/boost@lists.boost.org/msg09308.html is accurate, then it should be the same. It makes sense in the context.
8. Some more link updates. 

If you want me break this up into smaller commits or revert some of the changes, please let me know.

Kind regards,
Tinko Bartels

(Edit: I did not build the tests because I was not changing any actual code and because, in case I messed up, I am still trusting the CI to do that.)

### DIFF
--- a/doc/AStarHeuristic.html
+++ b/doc/AStarHeuristic.html
@@ -129,7 +129,7 @@ Called for the target of every out edge of a vertex being examined.
 <TABLE>
 <TR valign=top>
 <TD nowrap>Copyright &copy; 2004</TD><TD>
-<A HREF="http://www.cs.rpi.edu/~beevek/">Kristopher Beevers</A>,
+<A HREF="http://cs.krisbeevers.com/">Kristopher Beevers</A>,
 Rensselaer Polytechnic Institute (<A
 HREF="mailto:beevek@cs.rpi.edu">beevek@cs.rpi.edu</A>)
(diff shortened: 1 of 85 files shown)


anadon:
@jzmaddock Can we get this merged in quickly?  It looks good to me.

PR #169, opened 2019-04-26
Relocate boostorg/disjoint_sets into graph

### LINKED ISSUE #113: Consume boostorg/disjoint_sets
The repository disjoint_sets contains one non-detail header and it is in boost/pending.  A search of boost shows that only Boost.Graph and Boost.GraphParallel use disjoint_sets.  I would like to eliminate disjoint_sets as a separate repository to ease the burden on the CMT.  The maintainer of Boost.Container did not feel that disjoint_sets belonged there.  As such I would like to move this module into Boost.Graph and update Graph and GraphParallel appropriately.

jeking3:
Reasons for this:

1. The headers in boostorg/disjoint_sets have a dependency on graph, causing a cycle; this resolves the cycle.
2. Maintaining a repository for one class like disjoint_sets is unreasonable overhead.
3. The only consumers of disjoint_sets are graph and graph_parallel, and graph_parallel depends on graph.

The disjoint_sets headers are still in boost/pending, as graph already has a number of these.
The history of the files in disjoint_sets has not been preserved (I believe this would require a rebase and force push to achieve, and does not seem worth the trouble it would cause).

Completing this work will require some coordination - pull requests after this will be:

* Removing disjoint_sets from the superproject.
* Removing disjoint_sets from the boostorg exceptions. cc: @pdimov 

This may require an announcement for all developers so they can update their local sandboxes.

This fixes #113


jeking3:
It was suggested on the boost mailing list that it is acceptable for the builds to be broken during the migration, since our multirepo layout does not handle this type of consolidation well.  If you would consider merging this, I will submit a pull request into boostorg to remove disjoint_sets as a project.

anadon:
@jzmaddock Can we try this out?

jzmaddock:
Hopefully I've managed to coordinate these changes with disjoint_sets without breaking anything.... but we'll see!

Issue #167, opened 2019-04-16
labeled_graph::remove_vertex does not remove label


arneboe:
When calling `remove_vertex` on a `labeled_graph` the vertex is removed but the label is not.
A dangling reference to the vertex is left behind. Accessing it causes segfaults.

Minimal example:

```
#include <iostream>
#include <string>

#include <boost/graph/directed_graph.hpp>
#include <boost/graph/labeled_graph.hpp>

using namespace boost;
using namespace std;

int main() {

    using namespace boost::graph_detail;
    typedef directed_graph<> Digraph;
    typedef labeled_graph<Digraph, string> Graph;
    
    Graph g;
    g.add_vertex("foo");
    
    auto v = g.vertex("foo");
    if(v != nullptr)
        cout << "foo exists\n";
    
    g.remove_vertex("foo");
    
    auto v2 = g.vertex("foo");
    if(v2 != nullptr)
        std::cout << " BUG! vertex label still exists.\n";


    return 0;
}
```

anadon:
@jzmaddock Hey, this looks like a good one for me to work on.

jzmaddock:
Go for it!

anadon:
@arneboe Can I get your full name, since what you submitted is what I'm basing the regression test on?  I need to credit you.

arneboe:
I dont care about the credit, but if you have to: Arne Böckmann

anadon:
It looks like the intended design, given the state of BGL, is to have `_map` a protected member, not a private member.  So you hack was less of a hack than was expected.

anadon:
This may take longer than expected: https://github.com/boostorg/graph/blob/develop/include/boost/graph/adjacency_matrix.hpp#L992

anadon:
May be related: https://svn.boost.org/trac10/ticket/7863

anadon:
This is going to be delayed somewhat.  There are general cleanups in testing which I need to do first.

HWiese1980:
Sure, take your time! Better have it fixed correctly and thoroughly, than to rush it and only come up with some quick'n'dirty patch that fails somewhere else.

anadon:
Just an update, had to move twice, once across states, and started a new job.  Getting back to this more now.

HWiese1980:
Alright, no pressure. Just don't forget it ;-)

fjch1997:
And... it was forgotten
This bug really tripped me up when I tried to use labeled_graph

[...]

PR #170, opened 2019-04-30
Depreciation cleanups; require c++11

anadon:
@jzmaddock I have a bigger one for you.

Cleaned up a lot of old depreciation warnings, and starting to add logic for coverage CI.

Also, there were a few instances where C++11 was just so easier, and so now BGL requires C++11 support.  As I go forward, this will become more important.


anadon:
@Glenfe in Slack mentioned that some of the C++11 requirements w/r std::bind1st can be replaced with boost::bind which has C++03 support.  It looks like he's going to review a few of these changes.

include/boost/graph/labeled_graph.hpp
```diff
@@ -386,7 +386,10 @@ class labeled_graph
 
     /** Remove the vertex with the given label. */
     void remove_vertex(Label const& l)
-    { return boost::remove_vertex(vertex(l), _graph); }
+    {
+        // _map.erase(l);
```
anadon: Related to #167 .

jeremy-murphy:
As I have said previously, everyone appreciates effort, but I don't think this is the right way to progress the BGL. The best pull requests are small, not big; they focus on one thing, or one idea across several areas. This is a mixture of what looks like mostly white-space changes and actual code changes, which makes the actual code changes easy to miss. I recommend to just not worry about white-space at all. It's something we can fix automatically with clang-format one day -- no-one wants to review white-space changes, and it's just risky to mix them up with actual material changes.
Also, I don't think changing the base language standard of a library should happen on the whim of a pull request. I'm all for progress, but we also have a community that uses this library and they should probably be consulted with in case we happen to have a lot or a few significant users that rely on '03 compatibility.
Let's hope not, but maybe let's ask first?

anadon:
@jeremy-murphy Glen Fernandes caught me on a number of issues with this commit.  Check back in about a week or so.  I should have it in a shape you find considerably less objectionable.

.travis.yml
```diff
@@ -27,44 +27,6 @@ matrix:
...
-
-    - os: linux
-      env: TOOLSET=gcc COMPILER=g++-4.8 CXXSTD=03,11
-      addons:
-        apt:
-          packages:
-            - g++-4.8
-          sources:
-            - ubuntu-toolchain-r-test
-
```
jzmaddock: Boost in general is still testing with these older compiler, I'm not sure we want to remove all these?

example/fr_layout.cpp
```diff
@@ -16,7 +16,6 @@
 #include <map>
 #include <vector>
 #include <boost/random/linear_congruential.hpp>
-#include <boost/progress.hpp>
```
jzmaddock: I know boost/progress.hpp is old stuff, but I assume program output is a lot less pretty without this?  What's the rationale for removing it?

example/graph-assoc-types.cpp
```diff
@@ -10,23 +10,27 @@
...
 using namespace boost;
 
-template < typename Graph > void
-generic_foo(Graph & g)
-{
+template < typename Graph > 
+void
+generic_foo(
+  Graph & g
+){
```
jzmaddock: I know I said I'd accept any sensible reformating, but please not mixed up with substantive changes, and IMO this is a horrible format :(

example/graphviz.cpp
```diff
@@ -43,37 +42,41 @@ void test_graph_read_write(const std::string& filename)
...
 }
 
-int test_main(int argc, char* argv[])
-{
+
+int
+main(
+  int argc, 
+  char* argv[]
+){
```
jzmaddock: Sorry, but that formatting really does grate with me.

include/boost/graph/adjacency_list_io.hpp
```diff
@@ -76,7 +76,7 @@ void get
 }
 
 // get a subset of properties tagged Stag
-template<class Tag, class Value, class Next, 
+template<class Tag, class Value, class Next,
```
jzmaddock: Does this file contain anything other than whitespace changes?   It's *very* hard to review stuff like this.... sorry!

include/boost/graph/bc_clustering.hpp
```diff
@@ -147,14 +147,14 @@ betweenness_centrality_clustering(MutableGraph& g, Done done,
...
- */ 
+ */
 template<typename MutableGraph, typename Done>
 void
 betweenness_centrality_clustering(MutableGraph& g, Done done)
 {
   typedef typename Done::centrality_type centrality_type;
   std::vector<centrality_type> edge_centrality(num_edges(g));
-  betweenness_centrality_clustering(g, done, 
+  betweenness_centrality_clustering(g, done,
```
jzmaddock: Another 2 files with just whitespace changes: with over 200 changed files and a mix of substantive and formatting changes this is getting rather hard going.   Please can we split this into smaller orthogonal changes?

Stopping at this point...

anadon:
You can change the review settings to hide files with only whitespace
changes.

PR #173, opened 2019-06-10
adjacency_list: Fix example broken link

ffosilva:
The file was renamed from family-tree-eg.cpp to family_tree.cpp


anadon:
@jzmaddock Looks like this one is good to go.

jzmaddock:
Done.

PR #174, opened 2019-06-17
Update .travis.yml

CromwellEnage:
* Switch clang-linux test sources from llvm-toolchain-trusty-* to llvm-toolchain-xenial-*.
* Split some jobs that might otherwise exceed their time limits in the future.
* Fix name typo.


anadon:
I have a related branch in my repo which is incorporating work from boost-ci.  Think you could help on that one instead?

CromwellEnage:
Sure, want me to close this one?

PR #179, opened 2019-08-02
Fix build failure with gcc 4.9.1

tntljc:
- This is a gcc bug existing in 4.9.1 but it was fixed in 4.9.3. See comipler bug link https://gcc.gnu.org/bugzilla/show_bug.cgi?id=62255#c2 for details.


anadon:
I'm overhauling CI right now, so I'd like to put this merge on hold.

jeremy-murphy:
If it was fixed in 4.9.3 then I would say don't use anything less than that.

Issue #180, opened 2019-08-20
std:fill call in two_bit_color_map.hpp causes warning for downstream project in Visual Studio


jasjuang:
This [line](https://github.com/boostorg/graph/blob/develop/include/boost/graph/two_bit_color_map.hpp#L62) in `two_bit_color_map.hpp` I believe should be changed to 

```
std::fill(data.get(), data.get() + (n + elements_per_char - 1) / elements_per_char, (unsigned char)0);
```

If my downstream project happens to include `two_bit_color_map.hpp` I will have warning messages similar to the below, and it's because `data` is `unsigned char` yet it is being filled with 0, which is `int`.

```
1>mycpp.cpp
1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\xutility(2903): error C2220: warning treated as error - no 'object' file generated
1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\xutility(2917): note: see reference to function template instantiation 'void std::_Fill_unchecked1<_FwdIt,_Ty>(_FwdIt,_FwdIt,const _Ty &,std::false_type)' being compiled
1>        with
1>        [
1>            _FwdIt=unsigned char *,
1>            _Ty=int
1>        ]
1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\xutility(2925): note: see reference to function template instantiation 'void std::_Fill_unchecked<_Ty*,int>(_FwdIt,_FwdIt,const int &)' being compiled
1>        with
1>        [
1>            _Ty=unsigned char,
1>            _FwdIt=unsigned char *
1>        ]
1>C:\dev\vcpkg\installed\x64-windows-dev\include\boost/graph/two_bit_color_map.hpp(61): note: see reference to function template instantiation 'void std::fill<T*,int>(_FwdIt,_FwdIt,const _Ty &)' being compiled
1>        with
1>        [
1>            T=unsigned char,
1>            _FwdIt=unsigned char *,
1>            _Ty=int
1>        ]
1>C:\dev\vcpkg\installed\x64-windows-dev\include\boost/graph/two_bit_color_map.hpp(57): note: while compiling class template member function 'boost::two_bit_color_map<boost::vec_adj_list_vertex_id_map<Property,unsigned __int64>>::two_bit_color_map(size_t,const IndexMap &)'
[...]

PR #181, opened 2019-08-20
std::fill with char instead of int (fix issue #180)

jasjuang:
fix issue #180

### DIFF
--- a/include/boost/graph/two_bit_color_map.hpp
+++ b/include/boost/graph/two_bit_color_map.hpp
@@ -59,7 +59,7 @@ struct two_bit_color_map
     : n(n), index(index), data(new unsigned char[(n + elements_per_char - 1) / elements_per_char])
   {
     // Fill to white
-    std::fill(data.get(), data.get() + (n + elements_per_char - 1) / elements_per_char, 0);
+    std::fill(data.get(), data.get() + (n + elements_per_char - 1) / elements_per_char, '\0');
   }
 };
 


anadon:
I'd like to delay this until I get my CI PR finished.

glenfe:
You're replacing 0 with '0'? Why do you believe this is the solution?

jasjuang:
@glenfe It's not about believing it or not. This actually solves the compilation error as indicated in #180.

anadon:
@jasjuang I didn't look at this too closely, but understanding why it fixes it and why it is right is important.

jasjuang:
It is explained in previous threads. `data` is a `unsigned char`, `0` is an int, so there's an implicit type conversion when `std::fill` `data` with `0`, visual studio is strict hence it raises warnings against it. This PR turns `0` to `'0'`, so now it's filled with char and matches type with `data`, so visual studio stops complaining. Is this explanation clear enough?

anadon:
You didn't reference any threads here.  Furthermore, the only way to preserve identical behavior if what you are describing is complete is to use `'\0'`.  So it sounds like there is missing information.  What's up?

jasjuang:
Ok, I made the changes as requested. Nothing special is up, just trying to resolve the bug here.  Everything was explained in #180 and the link at the end of #180. Please merge. Thanks.

anadon:
@glenfe Looks clean.  Would you like to do the honors?

glenfe:
@jasjuang
>  It's not about believing it or not. This actually solves the compilation error as indicated in #180.

Just eliminating a compilation error in the wrong way is *not* a solution. '0' and 0 have two different values. 

If you want an `unsigned char` with a value of 0, you should use an unsigned char with a value of 0. Not one with a value of '0'.

Either `'\0'` or `staric_cast<unsigned char>(0)` would have been appropriate.

glenfe:
@anadon 
> Looks clean. Would you like to do the honors?

Instead of `new unsigned char[n]`  followed by `std::fill` to initialize all the elements to 0, I would prefer using just `new unsigned char[n]()` which would value-initialize all the elements to 0. This would eliminate the need to use `std::fill` entirely.

glenfe:
@anadon To elaborate on my comment above: Right now, the code is doing:

```
explicit two_bit_color_map(std::size_t n, const IndexMap& index = IndexMap())
  : n(n), index(index),
    data(new unsigned char[(n + elements_per_char - 1) / elements_per_char])
{
  // Fill to white
  std::fill(data.get(), data.get() + (n + elements_per_char - 1) / elements_per_char, 0);
}
```

I'm suggesting instead doing:

```
explicit two_bit_color_map(std::size_t n, const IndexMap& index = IndexMap())
  : n(n), index(index),
    data(new unsigned char[(n + elements_per_char - 1) / elements_per_char]())
{ }
```

Note the extra parenthesis `()` in the `new` expression for value-initialization, and the elimination of the `std::fill()` call.

anadon:
@glenfe I was just going through that and figuring out where you intended that.  I did find it before reading this!  I should have caught that.

glenfe:
See d3b157f4656a462beb9ca767838e69b60b393b4b

glenfe:
I've merged d3b157f to develop. We don't need this PR.

PR #186, opened 2019-09-28
Link-Cut-Tree implementation

yi-ji:
This PR implements Link-Cut-Tree data structure for Boost Graph Library. Since disjoint_sets algorithms has been moved into boost/graph (#169), this should also be the right place to put link_cut_tree algorithm.

This implementation of link_cut_tree supports following operations: 

- `make_tree(x)`: create a new link-cut-tree containing only root `x`
- `link(x, y)`: make the tree rooted at `x` a subtree of `y`;
- `cut(x)`: remove the edge connecting `x` to its parent;
- `find_root(x)`: find the root of the tree containing `x`;
- `lowest_common_ancestor(x, y)`: find the lowest common ancestor of such `x` and `y` that have same root.

All operations above are of amortized time complexity O(log n), except O(1) for `make_tree`.

This implementation refers to:
Robert Endre Tarjan, _Data Structures and Network Algorithms_, Chapter 5


jeremy-murphy:
I'm not sure that Boost.Graph really is the right place for `disjoint_sets`, but it definitely is the right place  for `link_cut_tree`!
I just have one question: how is the documentation going?  :)

include/boost/graph/link_cut_tree.hpp
```diff
@@ -0,0 +1,279 @@
...
+        inline Vertex lowest_common_ancestor(Vertex x, Vertex y)
+        {
+            return id_to_vertex[LCT::lowest_common_ancestor(get(id, x), get(id, y))];
+        }
+        
+    private:
+        ID id;
+        std::unordered_map<Index, Vertex> id_to_vertex;
+        IndexMapContainer parent_map, left_map, right_map;
+    };
```
jeremy-murphy: Did you consider parameterizing the associative container for `id_to_vertex`? Someone might have a use case where they can just use `std::vector` and they'll be awfully annoyed that `std::unordered_map` is hard-coded.

yi-ji: Thanks for advice, it's fixed. Also added documentation :)

yi-ji:
@jeremy-murphy comments resolved, checks passed :)

include/boost/graph/link_cut_trees.hpp
```diff
@@ -222,20 +222,21 @@ namespace boost
...
-    public link_cut_tree<associative_property_map<std::unordered_map<typename property_traits<ID>::value_type, typename property_traits<ID>::value_type> > >
+    template <class ID = identity_property_map, class InverseID = boost::unordered_map<typename property_traits<ID>::value_type, typename property_traits<ID>::key_type> >
+    class link_cut_trees_with_storage :
+    public link_cut_trees<associative_property_map<boost::unordered_map<typename property_traits<ID>::value_type, typename property_traits<ID>::value_type> > >
     {
     public:
         typedef typename property_traits<ID>::key_type Vertex;
         typedef typename property_traits<ID>::value_type Index;
-        typedef std::unordered_map<Index, Index> IndexMapContainer;
+        typedef boost::unordered_map<Index, Index> IndexMapContainer;
```
jeremy-murphy: This looks like another associative container type that should be parameterized?

yi-ji: Hi @jeremy-murphy Can we mark this as resolved or do you think it's necessary to parameterize it? :)

yi-ji:
Hi @jeremy-murphy  Thanks for reviewing! This PR has been ready for quite a while. Any concerns before merging it? Would appreciate someone with write access to have a look :) @jzmaddock

jeremy-murphy:
Hey, sorry, I don't have much time for open-source projects now that I have a baby! I'll try to review again quickly.

doc/bibliography.html
```diff
@@ -453,6 +453,11 @@ <H2>Bibliography</H2>
 <em>Data Structures for Weighted Matching and Nearest Common Ancestors with Linking</em><br>
 Proceedings of the First Annual ACM-SIAM Symposium on Discrete Algorithms, pp. 434-443, 1990.
 
+<p></p><dt><a name="tarjan83">77</a>
+<dd>Robert Endre Tarjan<br>
+<em>Linking and Cutting Trees</em><br>
+Data Structures and Network Algorithms, ISBN: 978-0-89871-187-5, pp. 59-70, 1983.
+
```
jeremy-murphy, line 460: Wow, I'm surprised that this is the first time that this book is referenced -- probably long overdue!

doc/link_cut_trees.html
```diff
@@ -0,0 +1,209 @@
...
+
+  <h3>Example</h3>
+
+  <p>A typical usage pattern for <tt>link_cut_trees</tt> can be seen in the
+  dynamic connectivity problem for acyclic graphs. Given two nodes <tt>x</tt> and <tt>y</tt>, 
+  they are connected if and only if <tt>find_root(x) == find_root(y)</tt>.</p>
+
+  <pre>
+  ...
+  link_cut_trees&lt;ElementParentMap, ElementChildMap&gt; lct(p, l, r);
```
jeremy-murphy, line 55: Where do those parameters--p, l, r--come from? I might be missing it, but I can't see them defined earlier.

include/boost/graph/link_cut_trees.hpp
jeremy-murphy: The problem is, the unordered containers in Boost and the standard library have pretty terrible performance characteristics. (Actually, I don't know that for a fact about Boost, I just assume that it is similar to `std`.)

So, if I were to use this algorithm, and I cared about performance, the _first_ thing I would do is replace `boost::unordered_map` with a different associative container.
And we can assume that people programming in C++ care about performance.
There is a pretty good talk about hash table performance here: https://www.youtube.com/watch?v=M2fKMP47slQ

So whilst we can't necessarily _provide_ our audience with the best hash table, we can at least _allow_ them to use one.
And even if we did provide it... a better one would come along later.

include/boost/graph/link_cut_trees.hpp
```diff
@@ -0,0 +1,280 @@
...
+
+namespace boost
+{
+    template<class ElementParentMap, class ElementChildMap = ElementParentMap>
+    class link_cut_trees
+    {
+    public:
+        inline link_cut_trees(ElementParentMap p, ElementChildMap l, ElementChildMap r) : parent(p), left(l), right(r) {}
+        
+        inline link_cut_trees(const link_cut_trees<ElementParentMap, ElementChildMap> &c)
```
jeremy-murphy, line 25: This is an unimportant topic but you might like to know that you can just use the class name (e.g. `link_cut_trees`) inside the class without the template parameters and it will automatically use the current values of them. It can help make code a little more readable.

include/boost/graph/link_cut_trees.hpp
```diff
@@ -0,0 +1,280 @@
...
+
+        template <class Element>
+        inline Element find_root(Element x)
+        {
+            return find_tail(expose(x));
+        }
+
+        template <class Element>
+        inline void link(Element x, Element y) // Element x must be a tree root
+        {
```
jeremy-murphy, line 42: Can you assert that x is a tree root so that making that mistake is caught during debugging? Maybe it's too difficult to test, I haven't actually looked into it yet.

jeremy-murphy: But looking at it again, maybe:
`BOOST_ASSERT(find_root(x) == x);`

include/boost/graph/link_cut_trees.hpp
```diff
@@ -0,0 +1,280 @@
...
+            expose(x);
+            std::pair<Element, Element> uv = split(x);
+            put_successor(x, x);
+            put_successor(uv.second, uv.second);
+        }
+
+        template <class Element>
+        inline Element lowest_common_ancestor(Element x, Element y) // Elements x and y must have same root
+        {
```
jeremy-murphy, line 59: Again, something like `BOOST_ASSERT(find_root(x) == find_root(y));`

include/boost/graph/link_cut_trees.hpp
jeremy-murphy: An afterthought: yeah, `disjoint_sets_with_storage` is hard-coded to use `std::vector`, and for a long time no-one would have challenged that. But now we, as a community, know that small-size optimizations can yield real benefits, so maybe even that data structure should become a parameter with `std::vector` as a default.

Interesting comparison to note, [`std::stack`](https://en.cppreference.com/w/cpp/container/stack) has exactly this kind of parameterization, but it defaults to `std::dequeue` for some strange reason.

include/boost/graph/link_cut_trees.hpp
yi-ji: Good to know! Fixed it.

doc/link_cut_trees.html
yi-ji: This code snippet is a tiny example showing the basic usage, with variables declaration/initiation omitted. Runnable examples can be found in `test/link_cut_trees_test.cpp`.
With that being said, I've updated their names to improve readability.

include/boost/graph/link_cut_trees.hpp
yi-ji: The amortized time complexity is _O(log n)_ for both `link` and `find_root`; However, if `find_root` is called inside each `link` call, then I'm afraid the amortized time complexity might not remain the same.

doc/bibliography.html
yi-ji: Thanks for helping review :smile: 
And enjoy parenting! :baby_bottle:

include/boost/graph/link_cut_trees.hpp
yi-ji: Makes sense to me, I agree. And I learned :+1: 
Done changing it as suggested :)

yi-ji:
@jeremy-murphy  Thanks for helping review again :smile: 
I've replied or made changes accordingly. 
Enjoy parenting! :baby_bottle:

include/boost/graph/link_cut_trees.hpp
jeremy-murphy: That's true, it will make the time complexity worse. In my opinion, that's OK in Debug mode, where a developer is concerned with correctness, not performance.
I definitely think this is a nice safeguard to have, but it's not critical.

Let's ask @jzmaddock what he thinks.

yi-ji: @jeremy-murphy I've just realized that `BOOST_ASSERT` can be optinally disabled using certain [macros](https://www.boost.org/doc/libs/1_67_0/libs/assert/doc/html/assert.html#boost_assert). So I've added the assertion as suggested :) If users would like to skip them for best performance, they can easily achieve that.

include/boost/graph/link_cut_trees.hpp
yi-ji: Done adding it.

yi-ji:
Hi @jeremy-murphy good sir, all comments resolved & checks passed, ready again :)

include/boost/graph/link_cut_trees.hpp
jeremy-murphy: Ah, OK, cool, that way everyone is happy.

yi-ji:
Can anyone help merge this?

jeremy-murphy:
I can't merge it, sorry, I don't have that access. @jzmaddock is the only person I'm aware of with that access and is active at the moment.

include/boost/graph/link_cut_trees.hpp
```diff
@@ -0,0 +1,283 @@
...
+        inline link_cut_trees(ElementParentMap p, ElementChildMap l, ElementChildMap r) : parent(p), left(l), right(r) {}
+        
+        inline link_cut_trees(const link_cut_trees &c)
+            : parent(c.parent), left(c.left), right(c.right) {}
+
+        template <class Element>
+        inline void make_tree(Element x)
+        {
+            make_path(x);
+        }
```
jeremy-murphy, line 32: Hey, I was just wondering... do you actually need to use the `inline` keyword on all these member functions? 
I'm not sure it has any effect, because I think they are all inline by virtue of being defined _in situ_.
It will just simplify reading the code a bit if we can remove them.

yi-ji: Yeah I agree, I have them removed now, thanks.
The CI builds are failing, but the error logs seem totally unrelated. 
Even a blank change failed to build: https://github.com/boostorg/graph/pull/243

jeremy-murphy: Dang, I can't help with the CI stuff, sorry.

jingpengw:
thanks for adding this!
It seems that this data structure could not be serialized using Boost: Serialization?

include/boost/graph/link_cut_trees.hpp
jeremy-murphy: Yeah, could be that the develop branch needs some cleaning up before anyone can merge new features.

yi-ji: No problem, CI builds passed now

yi-ji:
@jingpengw I guess it depends on the template parameters? All I know is that boost serialization supports [POD](https://en.cppreference.com/w/cpp/named_req/PODType) and STL containers, I think you can use them to instantiate link-cut trees.

yi-ji:
@jzmaddock Could you please take a look at this MR when you've got time? It's ready for merge :) Thank you.

jingpengw:
@jzmaddock did you get a chance to take a look? is it good to merge?

jeremy-murphy:
I need to get the CI fixed and I'm going to prioritize merging some bug fixes, but then one day I'll get back to you!  ;)

jingpengw:
any chance to merge this?

jeremy-murphy:
> any chance to merge this?

Yes, there is a chance.  :) I'll try to look over it again soon, but I can't guarantee anything.

jeremy-murphy:
> > any chance to merge this?
> 
> Yes, there is a chance. :) I'll try to look over it again soon, but I can't guarantee anything.

Btw, what I meant is, I definitely want to merge it, it's just a matter of chance that I find time to do it.  :)

include/boost/graph/link_cut_trees.hpp
```diff
@@ -0,0 +1,277 @@
...
+//
+#ifndef BOOST_LINK_CUT_TREES_HPP
+#define BOOST_LINK_CUT_TREES_HPP
+
+#include <boost/graph/properties.hpp>
+#include <boost/unordered_map.hpp>
+
+namespace boost
+{
+    template<class ElementParentMap, class ElementChildMap = ElementParentMap>
```
jeremy-murphy, line 19: I presume that `Element` in the Parent and Child maps have to be the same type, right? Or at the very minimum, they must be convertible to each other? The reason I ask is that I presume we can simplify the member functions from being templated on `Element` to being non-template functions. If we impose the strict requirement of being the same type then we can just get the element type from one of the map types, something like:
```
using Element = ElementParentMap::key_type;
```
Or if we want to allow more flexibility, then it might have to be:
```
using Element = typename std::common_type_t<ElementParentMap::key_type, ElementChildMap::key_type>;
```

What do you think?

include/boost/graph/link_cut_trees.hpp
```diff
@@ -0,0 +1,277 @@
...
+            return r;
+        }
+
+        template <class Element>
+        Element find_path(Element x)
+        {
+            splay(x);
+            return x;
+        }
+
```
jeremy-murphy, line 182: Not sure if this function needs to exist?

include/boost/graph/link_cut_trees.hpp
```diff
@@ -0,0 +1,277 @@
...
+            put(left, x, x);
+            put(right, x, x);
+            return std::make_pair<>(x_left, x_right);
+        }
+    };
+    
+    
+    template <class ID = identity_property_map, 
+        class InverseID = boost::unordered_map<typename property_traits<ID>::value_type, typename property_traits<ID>::key_type>, 
+        class IndexMapContainer = boost::unordered_map<typename property_traits<ID>::value_type, typename property_traits<ID>::value_type> >
```
jeremy-murphy, line 223: Boost.Unordered has a new, much faster, hash map called `unordered_flat_map`: https://www.boost.org/doc/libs/develop/libs/unordered/doc/html/unordered.html

It's not compliant with `unordered_map`, but I presume we don't need the extra features of the standard interface and can drop it in here.

Issue #188, opened 2019-10-03
Typo in name adjacenct_vertices()


AlekseyMuraviev:
I think it should be adjacent_vertices(), not adjacen**c**t_vertices().

In file https://github.com/boostorg/graph/blob/develop/include/boost/graph/labeled_graph.hpp.

```
/** @name Adjacency Graph */
//@{
template < LABELED_GRAPH_PARAMS >
inline std::pair< typename LABELED_GRAPH::adjacency_iterator,
    typename LABELED_GRAPH::adjacency_iterator >
adjacenct_vertices(
    typename LABELED_GRAPH::vertex_descriptor v, LABELED_GRAPH const& g)
{
    return adjacent_vertices(v, g.graph());
}
//@}
```

przemb:
I will change it :)

PR #189, opened 2019-10-05
Typo in name adjacenct_vertices()

przemb:
Hi,
This PR fixes typo reported in #188


jeremy-murphy:
So are there no graph concept checks for labeled graph? They would/should have caught this kind of typo. Could you add them, please?  :)

jeremy-murphy:
Fixed by PR #370.

Issue #175, opened 2019-07-12
std::max_element requires forward iterator


rogeeff:
std::max_element contract required forward iterator. Graph library's uses this interface in betweenness_centrality_clustering on edge iterators, which according to boost iterator library are deduced as input iterators. This fails to build with libc++, which validates the category at compile time. 

The easiest solution is to handcraft the for loop:

    auto it = edges_iters.first;
    auto it_max = it;
    while (++it != edges_iters.second) {
      if (cmp(*it_max, *it)) it_max = it;
    }
    // Here we assume graph is not empty and contains at least one edge.
    edge_descriptor e = *it_max;

anadon:
Hello,

I'm getting back to this repo after a month and a half.  While it is possible to make a loop in user code, that might not be the right option.  Though it may also be correct since a forward iterator may not make sense in a graph since it isn't convertible in a reliable way to a 1D directional structure.  I am working on Travis CI stuff right now and have a backlog, but I'll address this when I can.  If you would like a faster resolution and feel comfortable doing so, I will welcome you to create a test case and a code fix in a PR.

morinmorin:
C++20's new overloads of `std::max_element` would work. But, no need to wait for C++20! We already have `boost::range::max_element` (or `boost::first_max_element` in Boost.Algorithm) that works with ForwardTraversal ranges (iterators).

bkpoon:
I ran into this issue while trying to compile our code using Boost.Graph on macOS with Xcode 11. It looks like Boost.Range did what @morinmorin suggested by replacing `std::max_element` with `boost::first_max_element` (https://github.com/boostorg/range/commit/adcb071dc6b06e69aa4ddcc080d23695a5da378f).

For testing our code, I modified `include/boost/graph/bc_clustering.hpp` to add the `<boost/algorithm/minmax_element.hpp>` header, removed the `<algorithm>` header, and replaced line 136

            = *max_element(edges_iters.first, edges_iters.second, cmp);

with

            = *boost::first_max_element(edges_iters.first, edges_iters.second, cmp);

I'm still testing this on the other platforms and compilers that we support, though. I can submit a pull request if that helps.

bkpoon:
I made a pull request. There was another file, `include/boost/graph/howard_cycle_ratio.hpp`, that has the same issue, so I patched that as well.

Becheler:
Looks like #190 was merged and solved the bug, although this issue was never closed.

Replicate:

[...]

PR #190, opened 2019-10-16
std::max_element requires forward iterator

bkpoon:
1. Replace `std::max_element` with `boost::first_max_element` in `include/boost/graph/bc_clustering.hpp` as described in #175.
2. Additionally, `include/boost/graph/howard_cycle_ratio.hpp` has the same issue and is patched similarly in b9c3b14d3918e2beb25d2b8b4b735fc9c142cf7d.

I ran the basic tests. Everything compiles with Xcode 11.1 running on macOS 10.14.6 and it looks like nothing failed, but I'm not sure. I did have to remove `disjoint_sets.hpp` from `include/boost/pending` and `include/boost/pending/detail` for `b2` to work with the `develop` branch.

The testing is how I found the problem with `include/boost/graph/howard_cycle_ratio.hpp`. There is also a `std::max_element` in `include/boost/graph/planar_detail/bucket_sort.hpp`, but it looks like the arguments are acceptable iterators.
```
git clone https://github.com/boostorg/boost
cd boost
git submodule update --init

<change files in lib/graph>

./bootstrap.sh            <- compile b2
./b2 headers              <- just installs headers
./b2                      <- build compiled components

../../../b2               <- run all tests in lib/graph/test
```
Let me know if anything else is needed. Thanks!

### DIFF
--- a/include/boost/graph/bc_clustering.hpp
+++ b/include/boost/graph/bc_clustering.hpp
@@ -9,11 +9,11 @@
 #ifndef BOOST_GRAPH_BETWEENNESS_CENTRALITY_CLUSTERING_HPP
 #define BOOST_GRAPH_BETWEENNESS_CENTRALITY_CLUSTERING_HPP
 
+#include <boost/algorithm/minmax_element.hpp>
 #include <boost/graph/betweenness_centrality.hpp>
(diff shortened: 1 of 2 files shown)


bkpoon:
I changed the base from `master` to `develop`.

PR #204, opened 2020-02-13
Isomorphism improvements

### LINKED ISSUE #203: Isomorphism improvements
I want to improve the `isomorphism` documentation and discuss some improvements to the algorithm.

- Null vertices can appear in the isomorphism map if single-vertex components exist. This should either be explicitly documented or the algorithm adapted to map them, too.
- Vertex invariant functors actually require the `AdaptableUnaryFunction` concept, not `UnaryFunction` as stated (`argument_type` and `result_type` typedefs)
- In a named parameter call to `isomorphism`, the second vertex invariant functor additionally requires a `max()` member function as an alternative if `vertex_max_invariant` is not passed. ~~This is undocumented.~~ I think both `vertex_max_invariant` and `max()` here are misnomers, as it is demanded that they are upper **exclusive** bounds on the vertex invariant values (i.e. number of different values, not maximum values). Any added documentation of `max()` should have this information too, though.
- Both `vertex_max_invariant` and `vertex_invariant2::max()` are unnecessary: `test_isomorphism` immediately evaluates all invariants for all vertices of both graphs, sorts them, and lexicographically compares. The upper exclusive bound on the invariant values is then the maximum of the back of both invariant lists plus one. Yes, it is then less clear why invariants must be in a contiguous range with a reasonable upper bound - which, I assume, is purely for the efficiency and memory-compactness of the invariant multiplicity calculation here:
  ```c++
  // isomorphism.hpp, line 156 onwards
  std::vector<vertex1_t> V_mult;
    V_mult.push_back(v);
  {
    std::vector<size_type> multiplicity(max_invariant, 0);
      ++multiplicity.at(invariant1(v));
    sort(V_mult, compare_multiplicity(invariant1, &multiplicity[0]));
  }
  ```
  But I think that can be fixed with documentation, too. In principle, we could ignore the named parameter, even if supplied, and remove the ~~undocumented~~ requirement on `vertex_invariant2`, without breaking existing code.
- We could even go further and remove the range requirement on vertex invariants entirely. The multiplicity of invariants can be calculated efficiently by counting in the sorted list of invariants for one graph from the previous step. If the range requirement is removed, then it will need to be replaced by an `unordered_map` to avoid wasting memory.
- Something small: The first step of `test_isomorphism`, where all invariants are calculated for all vertices, does not `reserve`. 

Thoughts?

jargonzombies:
Closes #203 by
- Adding doc regarding null vertices in isomorphism maps when single-vertex components are involved
- Ignoring `max_vertex_invariant` arg and cheaply determining it in `test_isomorphism` (and adapting doc accordingly)
- Corrects `UnaryFunction` concept requirement in doc
- Reserves space in evaluation of all vertex invariants at start of `test_isomorphism`
- Removes range requirements on vertex invariants, determines invariant multiplicities from existing sorted list of all invariants using `unordered_map`
- Adds a colored isomorphism test
- Maps unmapped vertices if isomorphism passes

### DIFF
--- a/doc/isomorphism.html
+++ b/doc/isomorphism.html
@@ -92,8 +92,9 @@ href="./VertexListGraph.html">Vertex List Graph</a>.
 
 OUT: <tt>isomorphism_map(IsoMap f)</tt>
 <blockquote>
-The mapping from vertices in graph 1 to vertices in graph 2. This must
-be a <a href="../../property_map/doc/ReadWritePropertyMap.html">Read/Write
+The mapping from vertices in graph 1 to vertices in graph 2. May contain <a
+href="Graph.h">null vertices</a> if both graphs contain single-vertex
+components. This must be a <a href="../../property_map/doc/ReadWritePropertyMap.html">Read/Write
 Property Map</a>.<br> <b>Default:</b> an <a
 href="../../property_map/doc/iterator_property_map.html"><tt>iterator_property_map</tt></a>
 constructed from a <tt>std::vector</tt> of graph 2's vertex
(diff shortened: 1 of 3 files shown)


jeremy-murphy:
If you want each bullet point evaluated independently then it's best to make separate pull requests, otherwise it's all or nothing.
That would also help make clear which changes are simply fixing documentation for the existing implementation versus adding to/changing the implementation.

jargonzombies:
Hi Jeremy.

Thanks for the feedback. I've separated the points into #206 #207 and #208. Removing the range requirement on vertex invariants entirely is dependent on #206, so I will await the decision on it before creating that PR.

jeremy-murphy:
Thanks, I'll try to look at them soon.

jzmaddock:
Thanks @jargonzombies , I've been making some changes to this library, but don't claim to be any kind of Graph expert.

There is/was a plan to look at this lib and all open contributions during this years C++ Now conference, @JeffGarland is that still the plan?

PR #205, opened 2020-02-16
Isomorphism docs: AdaptableUnaryFn, null vertices

jargonzombies:
Isomorphism docs
- The expected concept for invariant functors is AdaptableUnaryFunction,
  not UnaryFunction
- Null vertices can appear in the isomorphism map output if both graphs
  contain disconnected vertices, add a note to that parameter's
  documentation

Related to #203

### DIFF
--- a/doc/isomorphism.html
+++ b/doc/isomorphism.html
@@ -92,8 +92,9 @@ href="./VertexListGraph.html">Vertex List Graph</a>.
 
 OUT: <tt>isomorphism_map(IsoMap f)</tt>
 <blockquote>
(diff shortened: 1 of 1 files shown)


jeremy-murphy:
Remind me, what does "Adaptable" mean here?

jargonzombies:
I had to check (it's been a while), but as far as I can tell it means (in addition to the `UnaryFunction` concept in `concept_check.hpp`) that the argument type is convertible to the concept function argument type.

jeremy-murphy:
Sorry about the long delay in acting on these, we just don't have many resources at the moment.

PR #206, opened 2020-02-16
Isomorphism: Ignore vertex_max_invariant

jargonzombies:
Isomorphism: Ignore `vertex_max_invariant`
- `vertex_max_invariant` and `invariant2.max()` are misnomers since what is
  expected is an upper exclusive bound on the possible invariant values,
  not their maximum value.
- The parameter can be ignored and the upper exclusive bound found
  cheaply at the start of `test_isomorphism`
- Removes the additional requirement of a nullary `max` member function on
  `invariant2`

Related to #203 

If this PR is accepted, then the range requirement on vertex invariants can be removed entirely with another PR.

### DIFF
--- a/doc/isomorphism.html
+++ b/doc/isomorphism.html
@@ -124,7 +124,8 @@ bound.
 
 IN: <tt>vertex_max_invariant(std::size_t max_invariant)</tt>
 <blockquote>
-This parameter is ignored.
+This parameter is ignored as it is no longer necessary, but kept for backwards
+compatibility.
 </blockquote>
 
 IN: <tt>vertex_index1_map(VertexIndex1Map i1_map)</tt>
--- a/doc/isomorphism.html
+++ b/doc/isomorphism.html
@@ -115,8 +115,8 @@ href="http://www.boost.org/sgi/stl/UnaryFunction.html">UnaryFunction</a>, with
 the argument type of <tt>vertex_invariant1</tt> being <tt>Graph1</tt>'s vertex
 descriptor type, the argument type of <tt>vertex_invariant2</tt> being
 <tt>Graph2</tt>'s vertex descriptor type, and both functions having integral
-result types.  The values returned by these two functions must be in the range
-[0, <tt>vertex_max_invariant</tt>).
+result types. The values returned by these two functions must have a low upper
+bound. 
 <br>
 <b>Default:</b> <tt>degree_vertex_invariant</tt> for both arguments<br>
 <b>Python</b>: Unsupported parameter.
@@ -124,14 +124,7 @@ result types.  The values returned by these two functions must be in the range
 
 IN: <tt>vertex_max_invariant(std::size_t max_invariant)</tt>
 <blockquote>
-An upper bound on the possible values returned from either
-vertex_invariant1 or vertex_invariant2.
-<br>
-<b>Default:</b> <tt>vertex_invariant2.max()</tt>. The default
-<tt>vertex_invariant2</tt> parameter, an instance of
-<tt>degree_vertex_invariant</tt>, defines this function to
-return <tt>num_vertices(g2) * (num_vertices(g2)+1)</tt>.<br>
-<b>Python</b>: Unsupported parameter.
+This parameter is ignored.
 </blockquote>
 
 IN: <tt>vertex_index1_map(VertexIndex1Map i1_map)</tt>
--- a/doc/isomorphism.html
+++ b/doc/isomorphism.html
@@ -116,7 +116,7 @@ the argument type of <tt>vertex_invariant1</tt> being <tt>Graph1</tt>'s vertex
 descriptor type, the argument type of <tt>vertex_invariant2</tt> being
 <tt>Graph2</tt>'s vertex descriptor type, and both functions having integral
 result types. The values returned by these two functions must have a low upper
-bound. 
+bound, as memory linear in the maximal invariant value is allocated.
 <br>
 <b>Default:</b> <tt>degree_vertex_invariant</tt> for both arguments<br>
 <b>Python</b>: Unsupported parameter.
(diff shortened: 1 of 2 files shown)


joemalle:
What do you think of using an unordered_map/similar instead of a vector to keep track of multiplicities?

jargonzombies:
> What do you think of using an unordered_map/similar instead of a vector to keep track of multiplicities?

It's been a while since I was read up on the algorithm, but if I'm not completely mistaken, I did exactly that in #204. Using an unordered map for the multiplicities was kind of the final point of that MR (I wanted to use a perfect hashing function for a vertex invariant), but I was advised to split that one into smaller bits for better reviewing. Unfortunately, no progress on any of them since.

Since nobody else seemed to have the expectation that the isomorphism would map disconnected vertices, I fixed it outside of the isomorphism call where I needed it. Seeing you face the same surprise, maybe we should be a bit more assertive about asking for review.

doc/isomorphism.html
```diff
@@ -115,23 +115,16 @@ <h3>Named Parameters</h3>
...
 <blockquote>
-An upper bound on the possible values returned from either
-vertex_invariant1 or vertex_invariant2.
-<br>
-<b>Default:</b> <tt>vertex_invariant2.max()</tt>. The default
-<tt>vertex_invariant2</tt> parameter, an instance of
-<tt>degree_vertex_invariant</tt>, defines this function to
-return <tt>num_vertices(g2) * (num_vertices(g2)+1)</tt>.<br>
-<b>Python</b>: Unsupported parameter.
+This parameter is ignored.
```
jeremy-murphy, line 127: Please just add a touch more detail to explain why it is ignored.

jeremy-murphy:
Again, sorry about the long wait times on review. It is a case of too many issues, not enough reviewers.

jeremy-murphy:
@joemalle , are you able to review this functionality?

include/boost/graph/isomorphism.hpp
```diff
@@ -165,18 +163,23 @@ namespace detail
             BGL_FORALL_VERTICES_T(v, G1, Graph1)
             f[v] = graph_traits< Graph2 >::null_vertex();
 
+            std::size_t max_invariant;
```
joemalle, line 166: This variable is used uninitialized.  `-Wall` would catch this.  Suggest using ` = 0;` or something along those lines

joemalle (changes requested):
This looks fine to me other than the initialization issue.  It would be nice to add a test

include/boost/graph/isomorphism.hpp
jargonzombies: I've added the initialization.

I do think you're incorrect, though. `max_invariant` is never read from before being assigned a value, and adding `-Wall` does not show any `-Wmaybe_unitialized` notices.

include/boost/graph/isomorphism.hpp
```diff
@@ -165,18 +163,26 @@ namespace detail
...
-                invar2_array.push_back(invariant2(v));
+                  invar2_array.push_back(invariant2(v));
                 sort(invar2_array);
                 if (!equal(invar1_array, invar2_array))
                     return false;
+
+                // Empty graphs case is covered before test_isomorphism is
+                // called, so the invar?_arrays cannot be empty, so back() is
+                // safe. Also the two invariant arrays are equal:
+                max_invariant = invar1_array.back() + 1;
```
jeremy-murphy, line 185: What if `invar1_array.back() == std::numeric_limits<std::size_t>::max()`?

It will happen eventually.

I had a quick look through the whole code for isomorphism but it's not obvious to me exactly what the simplest way of dealing with this is.

One option is that we only support values less than `std::numeric_limits<std::size_t>::max()` in the array and guard against it defensively using `assert`. I think this is reasonable.

jeremy-murphy: But I'm totally open to other suggestions.

jargonzombies: Well, there is the documentation change from

```diff
- The values returned by these two functions must be in the range [0, <tt>vertex_max_invariant</tt>).
+ The values returned by these two functions must have a low upper bound. 
```

The reason that's there is not the possibility of overflow, but rather the [invariant multiplicity counting](https://github.com/jargonzombies/graph/blob/ec286110269fc297bd3b3b6b687379b7d0afd8d3/include/boost/graph/isomorphism.hpp#L192), which allocates a zeroed vector of length equal to `max_invariant`, which is what I found when I encountered an out-of-memory error in trying a colored graph isomorphism with hashed invariants, prompting #203.

So to some degree I think adding an `assert` papers over a more fundamental issue that might want addressing.

The fix I would suggest is a [multiplicity counting using `unordered_map`](https://github.com/boostorg/graph/pull/204/files#diff-23655e4e06fd65a235692ad0debfc5c08c59dc6ba4cd4b27424c66df437b0bd8R177) instead, which eliminates `max_invariant` entirely, and allows consumers to use the full `size_t` range in their invariants without allocating excessive memory. Assuming this PR and #206 (which introduces `unordered_map` ) are accepted, I would have introduced another PR adding the remaining bits of #204, which, besides the counting, are just hashable concept checks on the invariant function result type and documentation changes.

I think there are two options:

- We add the multiplicity counting algorithm here.
- We add the assert and expand the documentation here, mentioning the allocation to explain the low bound documentation addition. If the follow-up PR with multiplicity counting gets accepted, that gets removed again.

I'm going to go with the latter for now. Let me know if you disagree.

jeremy-murphy: I'm inclined to trust your judgement here and go with whatever is the most efficient way for you to get it done. Sorry about wasting time with splitting the PR into multiple pieces, but at the time I felt like I wasn't going to be able to process it all at once.

jeremy-murphy:
Done, thank you!

include/boost/graph/isomorphism.hpp
jargonzombies: I do think it's better split up the way it is, so no worries. The original PR did do a lot of things at once.

PR #207, opened 2020-02-16
Isomorphism: Map disconnected vertices

jargonzombies:
Isomorphism: Map disconnected vertices
- Adds O(V) algorithm at `return_point_true` to match disconnected
  vertices in the isomorphism map
- Adds colored isomorphism test ensuring no null vertices are in the
  isomorphism map

Related to #203

### DIFF
--- a/include/boost/graph/isomorphism.hpp
+++ b/include/boost/graph/isomorphism.hpp
@@ -368,6 +368,8 @@ namespace detail
             {
             return_point_true:
                 {
+                    // At this point, there may still be null vertices in the 
+                    // mapping for disconnected vertices
                     std::vector< vertex1_t > unmatched_g1_vertices;
                     BGL_FORALL_VERTICES_T(v, G1, Graph1)
                     {
@@ -377,13 +379,13 @@ namespace detail
                     }
 
                     if(!unmatched_g1_vertices.empty()) {
-                        typedef unordered_multimap< invar2_value, vertex2_t > g2_invar_vertex_multimap;
-                        typedef typename g2_invar_vertex_multimap::iterator multimap_iter;
-                        g2_invar_vertex_multimap unmatched_invar_multimap;
+                        typedef unordered_multimap< invar2_value, vertex2_t > g2_invariant_vertex_multimap;
+                        typedef typename g2_invariant_vertex_multimap::iterator multimap_iter;
+                        g2_invariant_vertex_multimap unmatched_invariants;
                         BGL_FORALL_VERTICES_T(v, G2, Graph2)
                         {
                             if(!in_S[v]) {
-                                unmatched_invar_multimap.emplace(invariant2(v), v);
+                                unmatched_invariants.emplace(invariant2(v), v);
                             }
                         }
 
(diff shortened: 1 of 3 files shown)


joemalle:
Fixes #263

jargonzombies:
If #205 is merged before this, then this MR needs to be updated with removing the documentation addition noted there regarding possible null vertices in isomorphism maps

jeremy-murphy:
> If #205 is merged before this, then this MR needs to be updated with removing the documentation addition noted there regarding possible null vertices in isomorphism maps

Ahh, haha, I just merged #205 , so yes, sorry, we'll have to do that.

include/boost/graph/isomorphism.hpp
```diff
@@ -366,7 +367,39 @@ namespace detail
...
+                    {
+                        if(f[v] == graph_traits< Graph2 >::null_vertex()) {
+                            unmatched_g1_vertices.push_back(v);
+                        }
+                    }
+
+                    if(!unmatched_g1_vertices.empty()) {
+                        typedef unordered_multimap< invar2_value, vertex2_t > g2_invar_vertex_multimap;
+                        typedef typename g2_invar_vertex_multimap::iterator multimap_iter;
+                        g2_invar_vertex_multimap unmatched_invar_multimap;
```
jeremy-murphy, line 382: Please try not to repeat the type in the variable name, so in this case please remove the `_multimap` suffix.

Also, `invar` is a confusing contraction -- just use the full word `invariant` in all cases.

jeremy-murphy:
@joemalle , are you able to review this code too? It looks good to me, but it sounds like you have some experience with this functionality of the library, so your opinion would be appreciated.

joemalle:
I'm sorry, but I cannot help with this one.  I ran into this issue several years ago but haven't thought about it since.  The test looks convincing at the very least

jeremy-murphy:
@jargonzombies , did you need to make more changes here since the other PRs are merged or is this ready to go?

jargonzombies:
This is ready to go

include/boost/graph/isomorphism.hpp
```diff
@@ -366,7 +367,41 @@ namespace detail
...
+                        {
+                            invar1_value unmatched_g1_vertex_invariant = invariant1(*iter);
+                            multimap_iter matching_invariant = unmatched_invariants.find(unmatched_g1_vertex_invariant);
+                            BOOST_ASSERT(matching_invariant != unmatched_invariants.end());
+                            f[*iter] = matching_invariant->second;
+                            unmatched_invariants.erase(matching_invariant);
+                        }
+                    }
+                    return true;
+                }
```
jeremy-murphy, line 404: This `match` function is already huge, right, so I would prefer not to make it huger. Could you just put this bit of functionality into a private member function, please?

If you can abstract it to be a worthwhile non-member, generic function, that would be even better, but that could be a lot of work. Private member function would be a perfectly fine first step.

jeremy-murphy:
Thanks for the contribution and apologies again for the long delay getting it merged!

PR #234, opened 2020-12-13
Fix compilation on C++20 mode

jcelerier:
MSVC standard library has removed std::not1, thus use the boost alternative.

### DIFF
--- a/include/boost/graph/transitive_closure.hpp
+++ b/include/boost/graph/transitive_closure.hpp
@@ -158,9 +158,16 @@ void transitive_closure(const Graph& g, GraphTC& tc,
                 {
                     chain.push_back(v);
                     in_a_chain[v] = true;
+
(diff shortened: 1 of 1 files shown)


jcelerier:
@jzmaddock I'd appreciate a review, thanks !

jzmaddock:
This is fine, but introduces another dependency, I wonder if:

```
#ifdef __cpp_lib_not_fn
std::find_if(CG_vec[v].begin(), CG_vec[v].end(),
                            std::not_fn(detail::subscript(in_a_chain)));
#else
std::find_if(CG_vec[v].begin(), CG_vec[v].end(),
                            std::not1(detail::subscript(in_a_chain)));
#endif
```

Is any better?

jcelerier:
Oh right, seems like it's the correct replacement indeed. amending that

PR #235, opened 2020-12-16
maximum_weighted_matching: accomodate expression templates

sehe:
> weighted_augmented_path_finder with expression template

As discussed in https://stackoverflow.com/a/65331081/85371

Using a non-primitive weight type like

    using Weight = boost::multiprecision::cpp_dec_float_50;

Will run into errors because std::min fails to deduce with expression
templates. Disabling them:

    using Weight = boost::multiprecision::number<
            boost::multiprecision::cpp_dec_float<50>,
            boost::multiprecision::et_off >;

is, of course, not optimal. This change adds the explicit template
argument.

### DIFF
--- a/include/boost/graph/labeled_graph.hpp
+++ b/include/boost/graph/labeled_graph.hpp
@@ -784,7 +784,7 @@ inline typename LABELED_GRAPH::degree_size_type degree(
 template < LABELED_GRAPH_PARAMS >
 inline std::pair< typename LABELED_GRAPH::adjacency_iterator,
     typename LABELED_GRAPH::adjacency_iterator >
-adjacenct_vertices(
+adjacent_vertices(
     typename LABELED_GRAPH::vertex_descriptor v, LABELED_GRAPH const& g)
 {
     return adjacent_vertices(v, g.graph());
--- a/test/labeled_graph.cpp
+++ b/test/labeled_graph.cpp
@@ -4,23 +4,23 @@
 // Boost Software License, Version 1.0 (See accompanying file
 // LICENSE_1_0.txt or http://www.boost.org/LICENSE_1_0.txt)
 
-#include <iostream>
 #include <string>
-#include <set>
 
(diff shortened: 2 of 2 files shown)


jeremy-murphy:
Hmm, interesting. Thanks for identifying this issue. It would be great to have a test that actually uses an expression template type such as from Boost.Multiprecision. Would you mind adding it?

Are there any alternative ways of solving this worth considering?

Btw, I'm not going to merge any PRs until I fix the CI, so please be patient.

sehe:
The linked question contains a self-contained example [Workaround #3](http://coliru.stacked-crooked.com/a/f0ce705eb7299c81) using `maximum_weighted_matching`, which needs the patch in this PR to compile with 

```c++
using Weight = boost::multiprecision::cpp_dec_float_50;
```

As opposed to the version with expression templates disabled:

```c++
using Weight = boost::multiprecision::number<
        boost::multiprecision::cpp_dec_float<50>,
        boost::multiprecision::et_off >;
```

I don't think it's feasible to revisit all of BGL to make sure it all works fine with arbitrary precision number types (it's a combinatoric endeavour + coverage problem). However, this is a case of "low-hanging fruit" and perhaps it can serve as a nucleus/catalyst around which some more awareness can grow.

jeremy-murphy:
I want to merge a fix like this, but if there is no accompanying unit test that exercises it, then there is very little barrier to someone else breaking it again in the future. Also, a test case demonstrates that this change is complete and really does fix it -- everyone makes mistakes and forgets to include a complete change sometimes.

I strongly encourage everyone that contributes to Boost.Graph to follow a basic regime for contribution:
1. Add a test case that fails.
2. Add your change that makes it pass.

Otherwise, I have to do a lot of research to convince myself that a change is worthwhile. The test case makes it unambiguous.

I know you have a lot of experience and I trust your judgement, and I appreciate the effort you have already done, but that doesn't alleviate the concerns I raised.

PR #252, opened 2021-04-20
Added eigenvector centrality

Amritha16:


Valliammai19:
Eigenvector centrality is a useful centrality measure. Although Boost has its directed variant, Pagerank @Amritha16 and I couldn't find an implementation of Eigenvector centrality

jeremy-murphy:
Thanks for contributing this idea. The two algorithms look almost identical. Rather than duplicate the code with a minor change, I think it would be much better to rename the `page_rank` code to a general name, node_rank or whatever, parameterize the it so that then `page_rank` and `eigenvector_centrality` just use `node_rank`  with a functor that calculates that `u_rank_out`.
How does that sound?

Some unit tests wouldn't go astray either.  ;)

Amritha16:
@jeremy-murphy That sounds much better than our initial commit. Thanks for the suggestion!:)
@Valliammai19 and I will get on this next!

PR #257, opened 2021-05-05
Update adjacency_list.hpp

shrek1402:
If the project has a function with the same name, a name conflict occurs
error C2668: add_edge


jeremy-murphy:
Hmmm, I think it's intentionally unqualified so that the function is found via ADL, although I could be wrong.
What was the signature of your `add_edge` function and what namespace was it in?

Issue #258, opened 2021-05-13
BOOST_GRAPH_MPI_INCLUDE formatting


cojuer:
When trying to build parallel graph examples using boost 1.76 I get errors like this:
```
In file included from /usr/local/include/boost/graph/breadth_first_search.hpp:26,
                 from src.cpp:19:
/usr/local/include/boost/graph/two_bit_color_map.hpp:106:34: fatal error:  boost / graph / distributed / two_bit_color_map.hpp: No such file or directory
  106 | #include BOOST_GRAPH_MPI_INCLUDE(< boost / graph / distributed / two_bit_color_map.hpp >)
      |
```
Seems to be caused  by new formatting introduced after 1.72 when: 
`BOOST_GRAPH_MPI_INCLUDE(<boost/graph/distributed/two_bit_color_map.hpp>)` 
was replaced with 
`BOOST_GRAPH_MPI_INCLUDE(< boost / graph / distributed / two_bit_color_map.hpp >)`

jeremy-murphy:
Meaning I can close this?

PR #259, opened 2021-05-16
Fix formatting in BOOST_GRAPH_MPI_INCLUDE macro

cojuer:
Issue #258

### DIFF
--- a/include/boost/graph/breadth_first_search.hpp
+++ b/include/boost/graph/breadth_first_search.hpp
@@ -27,7 +27,7 @@
 #include <boost/graph/detail/mpi_include.hpp>
 #include <boost/concept/assert.hpp>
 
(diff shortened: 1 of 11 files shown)


andreacassioli:
I believe this is no more an issue and it can be closed.

andreacassioli:
@jeremy-murphy FYI

jeremy-murphy:
@andreacassioli but then why do the file changes still look valid?

cojuer:
@jeremy-murphy Probably because changes here and in commit `03318bf` (was created later than this one) lead to the same results and git wil just skip this commit if it is applied. And it looks like the issue was really fixed.

jeremy-murphy:
I don't have time to confirm, so I'll take your words on it! Thanks for all your efforts.

PR #264, opened 2021-06-02
Resolves #261 by replacing child() with first_child()

### LINKED ISSUE #261: Redundant `child_idx` argument of `d_ary_heap_indirect<>::child()` method
[child](https://github.com/boostorg/graph/blob/d865589d62afba7ca590bfcaf76f0723f8494dd4/include/boost/graph/detail/d_ary_heap.hpp#L208..L211) method takes `child_idx` as an argument.
```cpp
static size_type child(size_type index, std::size_t child_idx)
{
    return index * Arity + child_idx + 1;
}
```
But is never called with indices other than [0](https://github.com/boostorg/graph/blob/d865589d62afba7ca590bfcaf76f0723f8494dd4/include/boost/graph/detail/d_ary_heap.hpp#L307).
```cpp
size_type first_child_index = child(index, 0);
```
So why not get rid of it altogether, as in [other implementations](https://github.com/dotnet/runtime/blob/d3e35e7fd3dad543c275d1f1e10fc0ba91d70a15/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs#L586)?
```cs
private int GetFirstChildIndex(int index) => (index << Log2Arity) + 1;
```

qbit86:


jeremy-murphy:
I'd personally prefer to depend on Boost.Heap and use `d_ary_heap` from there. Not sure why it's currently copied into this detail directory.

qbit86:
> I'd personally prefer to depend on Boost.Heap and use `d_ary_heap` from there. Not sure why it's currently copied into this detail directory.

I don't believe that specific `d_ary_heap_indirect<>` could be easily replaced with general purpose `d_ary_heap<>`.
The former manipulates values indirectly via explicit indices and has some hacks [1] to cooperate with external logic of updating the underlying map.

[1] [boost/graph/detail/d_ary_heap.hpp#L161..L164](https://github.com/boostorg/graph/blob/b76ffc7f2a56955e2cdeb93f6902c40d3bbcffdd/include/boost/graph/detail/d_ary_heap.hpp#L161..L164)

jeremy-murphy:
Oh, dear, that's a bit disappointing.

jeremy-murphy:
Thanks for this!

PR #265, opened 2021-06-03
Resolves #260 by removing call to `combine`

### LINKED ISSUE #260: Confusing call to `combine` in `astar_search_no_init_tree`
What does this line of code mean in [astar_search.hpp](https://github.com/boostorg/graph/blob/d865589d62afba7ca590bfcaf76f0723f8494dd4/include/boost/graph/astar_search.hpp#L331)?
```cpp
combine(get(distance, v), e_weight);
```
The result of this `combine()` is not used anywhere. Actual distance update happens inside the `relax()` function.

qbit86:


jeremy-murphy:
Have you had a chance to step through this with the debugger and convince yourself that it truly has no effect?

qbit86:
> Have you had a chance to step through this with the debugger and convince yourself that it truly has no effect?

This function is provided by the user side as a part of monoid contract. It can have any observable effects as the user desires. Usually it is just the pure function adding two numbers. But of course it can format the user's disk or whatever she wants.

So actual effects are not the point of this issue. The point is that this call makes no sense in the context of A-star algorithm.

PR #267, opened 2021-06-09
Convert the few remaining tests to use the header-only Lightweight Test

glenfe:
Most of the tests in Graph already use LWT; this just converts the handful that don't.


jzmaddock:
Good idea, failures are all unrelated (Boost.Filesystem) so merging.

PR #269, opened 2021-07-02
Revert change to `reindex_edge_list`.

### LINKED ISSUE #268: Segfault for `remove_vertex` with `adjacency_list<boost::setS, ...>`
## Bug description
Removing a vertex in an `adjacency_list` graph produces an invalid read when the following conditions are met
- the edge storage `OutEdgeList` is `setS` or `hash_setS`
- the vertex storage `VertexList` is `vecS`
- a source or target of an edge in the graph has index larger than the vertex index being deleted

## Reproducing MWE
The following code:
```

using Graph = boost::adjacency_list<boost::setS, boost::vecS, boost::undirectedS>;

int main() {
  while (true) {
    Graph g(3);
    boost::add_edge(2, 0, g);
    boost::remove_vertex(1, g);
  }
}
```

On MacOS Big Sur, this results in a segfault in roughly 1/1000 iterations. This could be reproduced both with `AppleClang 12.0.5.12050022` and `LLVM Clang 12.0.0`. On Linux with gcc, no segfault was observed but the memory corruption can be observed in Valgrind (see [output](https://github.com/boostorg/graph/files/6748470/valgrind.txt)).

## Proposed fix
Reverting commit [5dd748307cc35e8](https://github.com/boostorg/graph/commit/5dd748307cc35e8ee1e7045902ba08adedbe8646#diff-406d7d1b8a0593dc27a38ce319[%E2%80%A6]85f11fb22a91f2d0326d19e8b27746d9) seems to fix the issue. The issue seems to come from incrementing an invalid iterator after a change to the original container.

cqc-alec:
Fixes #268  .

### DIFF
--- a/include/boost/graph/detail/adjacency_list.hpp
+++ b/include/boost/graph/detail/adjacency_list.hpp
@@ -2070,7 +2070,7 @@ namespace detail
         EdgeList& el, vertex_descriptor u, boost::disallow_parallel_edge_tag)
     {
         typename EdgeList::iterator ei = el.begin(), e_end = el.end();
-        while (ei != e_end) {
+        while (ei != e_end)
         {
             if (ei->get_target() > u)
             {
--- a/include/boost/graph/detail/adjacency_list.hpp
+++ b/include/boost/graph/detail/adjacency_list.hpp
@@ -2069,14 +2069,19 @@ namespace detail
     inline void reindex_edge_list(
         EdgeList& el, vertex_descriptor u, boost::disallow_parallel_edge_tag)
     {
-        for (typename EdgeList::iterator ei = el.begin(); ei != el.end(); ++ei)
+        typename EdgeList::iterator ei = el.begin(), e_end = el.end();
+        while (ei != e_end) {
         {
             if (ei->get_target() > u)
             {
                 typename EdgeList::value_type ce = *ei;
+                ++ei;
                 el.erase(ce);
                 --ce.get_target();
                 el.insert(ce);
+            }
+            else {
+                ++ei;
             }
         }
     }
--- a/test/delete_edge.cpp
+++ b/test/delete_edge.cpp
@@ -55,5 +55,13 @@ int main(int argc, char* argv[])
 
     BOOST_TEST(m_graph[ed1] == EDGE_VAL);
 
+    // https://github.com/boostorg/graph/issues/268
+    {
+        boost::adjacency_list<boost::setS, boost::vecS, boost::undirectedS> g;
+        boost::add_edge(2, 0, g);
+        boost::remove_vertex(1, g);
+    }
+
+
     return boost::report_errors();
 }
--- a/test/delete_edge.cpp
+++ b/test/delete_edge.cpp
@@ -56,12 +56,22 @@ int main(int argc, char* argv[])
     BOOST_TEST(m_graph[ed1] == EDGE_VAL);
 
     // https://github.com/boostorg/graph/issues/268
+    // 1. Undirected:
     {
         boost::adjacency_list<boost::setS, boost::vecS, boost::undirectedS> g;
         boost::add_edge(2, 0, g);
         boost::remove_vertex(1, g);
+        BOOST_TEST(num_vertices(g) == 2);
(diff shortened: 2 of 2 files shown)


include/boost/graph/detail/adjacency_list.hpp
```diff
@@ -2069,15 +2069,20 @@ namespace detail
...
     {
-        for (typename EdgeList::iterator ei = el.begin(); ei != el.end(); ++ei)
+        typename EdgeList::iterator ei = el.begin(), e_end = el.end();
+        while (ei != e_end)
         {
             if (ei->get_target() > u)
             {
                 typename EdgeList::value_type ce = *ei;
+                ++ei;
                 el.erase(ce);
```
prince-chrismc, line 2079:
```suggestion
                ei = el.erase(ce);
```

prince-chrismc (commented):
You should be able to keep the for loop and remove the increments if you correctly update after the erease

include/boost/graph/detail/adjacency_list.hpp
cqc-alec: Thanks for the suggestion. It indeed looks like a better approach; however, I was not able to get this code to compile:

```

using Graph = boost::adjacency_list<boost::setS, boost::vecS, boost::undirectedS>;

int main() {
  Graph g(3);
  boost::add_edge(2, 0, g);
  boost::remove_vertex(1, g);
}
```

The gist of the error is that `erase` does not return an iterator in this case, but a `long unsigned int`.

(The original patch was simply a reversion of a previous commit.)

jeremy-murphy:
Could you please add the example from the bug report that can trigger this segfault as a new unit test?
As long as it is reliable under valgrind or a memory sanitizer, that is sufficient.

cqc-alec:
> Could you please add the example from the bug report that can trigger this segfault as a new unit test? As long as it is reliable under valgrind or a memory sanitizer, that is sufficient.

Yes, I will do.

cqc-alec:
I have added the unit test to `delete_edge.cpp`, which seems the most appropriate existing test file -- is that OK or would you prefer a new source file?

jeremy-murphy:
> I have added the unit test to `delete_edge.cpp`, which seems the most appropriate existing test file -- is that OK or would you prefer a new source file?

I think the location is fine.

Is there a reason that you can't use the existing `m_graph` object? I know the MWE given uses undirected edges but the explanation does not stipulate it. Hmm. Anyway, it's probably good to test on both directed and undirected.

I know that we're primarily testing for a segfault, but it looks weird in a test if there is no accompanying assertion, so please add in a BOOST_TEST for i) the expected number of vertices and ii) edges after calling `remove_vertex`.

Thank you!!

jeremy-murphy:
Looks great, thanks. As soon as the CI is fixed, I'll merge this.

jeremy-murphy:
Please update to the latest changes on `develop` so we can test for a green build.

cqc-alec:
> Please update to the latest changes on `develop` so we can test for a green build.

Done. Still seems to be one build failure (looks unrelated) ...?

jeremy-murphy:
> > Please update to the latest changes on `develop` so we can test for a green build.
> 
> Done. Still seems to be one build failure (looks unrelated) ...?

Yeah, that is strange and annoying.

jeremy-murphy:
Might be a real issue with Boost.Filesystem, let's see: https://github.com/boostorg/filesystem/issues/227

jeremy-murphy:
The simplest way I know to re-run the CI is to close the issue and open it again!

jeremy-murphy:
Hmmm, that didn't work as I expected.

jeremy-murphy:
A couple of issues have been fixed so I think if you merge develop now, you should get a green build.

cqc-alec:
> A couple of issues have been fixed so I think if you merge develop now, you should get a green build.

Done: seems good now!

jeremy-murphy:
Awesome!

In the future, I think it's good to refer to the commit that you're reverting (I put it in the merge commit), so that people can easily track the history.

PR #280, opened 2021-12-14
Add user callback to `vf2_subgraph_mono` to potentially stop search before each check.

cqc-alec:
Also update documentation and add tests.

See the discussion on #271 .

Remarks:

* I used a simple function pointer for the type rather than a templated "callable" so as to be able to default it, thereby maintaining compatibility with existing code. An alternative would be to define a new method or methods with a different name; this seems a bit like overkill but I can do that if preferred.
* I only changed `vf2_subgraph_mono`, not `vf2_subgraph_iso`, but it should be simple to do the same thing there too if desired.
* The tests are pretty basic. It isn't obvious to me how to test this fully without knowing internal details of the order of search. Suggestions welcome.

### DIFF
--- a/doc/vf2_sub_graph_iso.html
+++ b/doc/vf2_sub_graph_iso.html
@@ -44,7 +44,8 @@ template &lt;typename GraphSmall,
           typename SubGraphIsoMapCallback&gt;
 bool vf2_subgraph_iso(const GraphSmall&amp; graph_small,
                       const GraphLarge&amp; graph_large,
-                      SubGraphIsoMapCallback user_callback)
+                      SubGraphIsoMapCallback user_callback,
+                      bool(*user_step_callback)() = &vf2_trivial_step_callback)
 
 
 <em class="comment">// Named parameter version</em>
@@ -59,7 +60,8 @@ bool vf2_subgraph_iso(const GraphSmall&amp; graph_small,
                       const GraphLarge&amp; graph_large,
                       SubGraphIsoMapCallback user_callback,
                       const VertexOrderSmall&amp; vertex_order_small,
-                      const bgl_named_params&lt;Param, Tag, Rest&gt;&amp; params)
+                      const bgl_named_params&lt;Param, Tag, Rest&gt;&amp; params,
+                      bool(*user_step_callback)() = &vf2_trivial_step_callback)
 
 
 <em class="comment">// Non-named parameter version</em>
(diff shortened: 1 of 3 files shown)


doc/vf2_sub_graph_iso.html
```diff
@@ -44,7 +44,8 @@ <h1>
           typename SubGraphIsoMapCallback&gt;
 bool vf2_subgraph_iso(const GraphSmall&amp; graph_small,
                       const GraphLarge&amp; graph_large,
-                      SubGraphIsoMapCallback user_callback)
+                      SubGraphIsoMapCallback user_callback,
+                      bool(*user_step_callback)() = &vf2_trivial_step_callback)
```
jeremy-murphy, line 48: Instead of being a concrete function pointer, this should be a template like most other parameters. That way, users can pass in whatever they like: lambda, function object, etc.

include/boost/graph/vf2_sub_graph_iso.hpp
```diff
@@ -773,6 +779,10 @@ namespace detail
         bool found_match = false;
 
     recur:
+        if (!user_step_callback()) {
```
jeremy-murphy, line 782: I think this callback should take in whatever data might change between each call to it. Even though in your case you'll ignore it, someone else might have a reason to examine the internals of whatever state is changing. I guess it's just that `s` object?

jeremy-murphy:
Looks great so far! Thanks for keeping it simple, sorry it took me so long to take a look. I just made a couple of requests for changes.

cqc-alec:
Thank you for the review Jeremy! I have actually switched to a different solution now, so no longer need this. I may come back to it anyway if you or others think it generally useful, though I cannot prioritize it just now. So, feel free to close, complete or leave open this PR as you see fit.

By the way, the reason I didn't use a template type for the callback was that I couldn't see how to do so while maintaining backward compatibility (i.e. how to set a no-op default value). I guess an alternative would be to overload the whole function.

jeremy-murphy:
OK, no worries. I'll leave this open for a while in case someone else needs the same functionality and wants to finish it off.

cqc-alec:
I wonder if it's time to close this now @jeremy-murphy ?

cqc-alec:
> > I wonder if it's time to close this now @jeremy-murphy ?
> 
> I still think it's a good idea, so I'll finish it off one day if you don't. :)

OK!

Becheler:
Realted to RFC #494 
Thanks @andreacassioli  for crossing it over 😄

Issue #272, opened 2021-07-19
Karp's minimum mean cycle algorithm


ggleizer:
Dear all,

Are there plans to implement [Karp's minimum mean cycle ](https://www.sciencedirect.com/science/article/pii/0012365X78900110) algorithm in Boost? If not, I would be willing to contribute (although my C++ is quite rusty). 

An example implementation is available [here](https://www.geeksforgeeks.org/karps-minimum-mean-average-weight-cycle-algorithm/). I would include recovering a minimizing cycle in the function; the added cost is small, because the main algorithm runs in O(mn) time (m edges and n vertices), while recovering the minimizer is O(n). For reference, Karp's original paper has a mistake in what concerns recovering the cycle, which was corrected by [Chatuverdi and McConnel](https://www.sciencedirect.com/science/article/pii/S0020019017301084?casa_token=NO-MxsCiH54AAAAA:cGZkyJwDDRJllboC2Wm7mDz2JX1xjJFw5Us9Ef_u9SXIEcY99Zx9DDNc2zCGK2cuaUMhR3PQxA).

Thank you,

Gabriel.

jeremy-murphy:
If there is no existing issue or pull request open for it then presumably no-one else is working on it.
Sounds like a good contribution to Boost.Graph. You don't have to be a C++ wizard to contribute, just be prepared for a lot of _feedback_. :)
The only problem is that the library is in a bit of a dormant state -- it's failing the CI and I don't know if any contributions will get merged until it's passing again.

ggleizer:
Thanks, Jeremy. I will (slowly) give it a try then.

GYuvanShankar:
Greetings,

I have tried my implementation of this algorithm  at https://github.com/GYuvanShankar/boost_graph_algorithm/blob/master/main.cpp. 
Please do check it out, and let me know if I can continue to open a pull request on this matter.

Thank you.
G Yuvan Shankar

Becheler:
Hi @ggleizer , @GYuvanShankar !
This issue recently came to my attention, and I would be willing to help you getting this merged. Are you still around ? 
Thanks again for your work 🙏🏽

ggleizer:
Hi @Becheler,

Sorry I have disappeared ever since offering to help; I had also missed @GYuvanShankar's message. From my perspective, I would be very happy with your support. Looking at @GYuvanShankar's code, it seems it does not return a minimizing cycle. Do you think it would be good to include it?

Thanks a lot!

[...]

PR #281, opened 2021-12-19
Added Karps Minimum Cycle Algorithm

GYuvanShankar:
Based on issue #272.
Implemented [Karps Minimum Cycle Algorithm](https://www.sciencedirect.com/science/article/pii/0012365X78900110).


jeremy-murphy:
Thanks for taking the time to implement this. It looks good, but as I said earlier it will take a few rounds of feedback to get it to be consistent with existing BGL code.

Please take a look at an existing similar algorithm such as [Howard's cycle ratio algorithm](https://www.boost.org/doc/libs/1_78_0/libs/graph/doc/howard_cycle_ratio.html)  and imitate what is there. There is a lot more to an algorithm than just the implementation, namely:
- Documentation
- Examples
- Unit tests

jeremy-murphy:
I marked this as a draft for now in the hope that it won't trigger the CI every time you push -- I'm still learning how this CI system works. When you have completed everything then mark it ready for review again.

jeremy-murphy:
Although one thing that I do not expect anyone to imitate from the existing code are named parameters.

include/boost/graph/karp_minimum_mean_cycle.hpp
```diff
@@ -0,0 +1,79 @@
...
+#define BOOST_GRAPH_KARP_MINIMUM_MEAN_CYCLE_HPP
+
+#include <iostream>
+#include <boost/graph/graph_traits.hpp>
+#include <boost/graph/adjacency_list.hpp>
+
+namespace boost
+{
+    template<typename Graph>
+    double karp_minimum_mean_cycle(Graph g)
```
jeremy-murphy, line 11: Although it doesn't roll off the tongue very well, the paper does call it the "minimum cycle mean".

GYuvanShankar:
Thanks for looking into this. I will start the documentation, was just waiting for someone to approve of the code.

jeremy-murphy:
Hi @GYuvanShankar , thanks, it's looking good! I'll start a review, in which I'll make some (or many) small comments in the code itself.

jeremy-murphy (changes requested):
I'm so sorry, I didn't realize that these review comments were in a "pending" state, not published.

jesseli2002:
> PS. One thing that I do not expect anyone to imitate from the existing code are named parameters.
Why not? Are named parameters being deprecated or something?

jeremy-murphy:
> > PS. One thing that I do not expect anyone to imitate from the existing code are named parameters.
> > Why not? Are named parameters being deprecated or something?

It was a courageous and ambitious experiment at the time (20 something years ago) but it adds an awful lot of complexity and maintenance burden to the code for questionable benefit. In the long run, I would like to strip them out. I think we could achieve similar benefits, if wanted, from much simpler overloads and stronger types.

PR #302, opened 2022-07-17
Fix conversion warning

### LINKED ISSUE #293: -Wconversion warning in transitive_closure.hpp
https://github.com/boostorg/graph/blob/c583bfa197ba324806dfb10db99c76daa8ff2fd0/include/boost/graph/transitive_closure.hpp#L85

`strong_components` might return something larger than an `int` can hold. This should probably use the corresponding trait (`vertices_size_type`?) instead.

sebrockm:
Fixes #293 

Replaces `int` with the type returned by `strong_components` which is the value type of the provided component map.
Also removes unneeded `erase` call.

### DIFF
--- a/include/boost/graph/transitive_closure.hpp
+++ b/include/boost/graph/transitive_closure.hpp
@@ -107,12 +107,11 @@ void transitive_closure(const Graph& g, GraphTC& tc,
             }
         }
         std::sort(adj.begin(), adj.end());
-        typename std::vector< cg_vertex >::iterator di
+        const typename std::vector< cg_vertex >::iterator di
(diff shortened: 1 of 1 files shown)


jeremy-murphy:
Hi @sebrockm , this looks great, thanks. Good catch of the redundant `.erase` -- I think a lot of people don't realize that it's an unnecessary step!

One thing -- could you please revert the Clang format commit? I really prefer to keep non-functional changes like formatting out of pull requests completely.

jeremy-murphy:
> Would a formatting-only PR make sense? It looks like there are some more files that are not fully in line with clang-format.

One day, yes, but it's just not a priority at the moment.

PR #310, opened 2022-10-21
Add tests for the all defaults overload of `boykov_kolmogorov_max_flow`

### LINKED ISSUE #309: The 3-parameter overload of `boykov_kolmogorov_max_flow` does not compile
According to the [documentation](https://www.boost.org/doc/libs/1_80_0/libs/graph/doc/boykov_kolmogorov_max_flow.html), there is a minimalistic overload of `boykov_kolmogorov_max_flow` that only takes the graph, a source, and a sink. All other parameters are set to their defaults (as stated in the section "Named Parameters").

However, this overload doesn't compile. Minimal reproducible example:

``` C++
#include <boost/graph/graph_traits.hpp>
#include <boost/graph/properties.hpp>
#include <boost/graph/boykov_kolmogorov_max_flow.hpp>


#include <utility>


namespace my_graph
{

struct Graph
{
    // graph concept
    using vertex_descriptor = size_t;
    using edge_descriptor = std::pair<size_t, size_t>;
    using directed_category = boost::directed_tag;
    using edge_parallel_category = boost::disallow_parallel_edge_tag;
    struct traversal_category : boost::vertex_list_graph_tag,
                                boost::edge_list_graph_tag,
                                boost::incidence_graph_tag,
                                boost::adjacency_matrix_tag
    {
    };

    static vertex_descriptor null_vertex() { return 0; }

    // incidence graph concept
    using out_edge_iterator = std::nullptr_t;
    using degree_size_type = size_t;

    // vertex list graph concept
    using vertex_iterator = std::nullptr_t;
    using vertices_size_type = size_t;

    // edge list graph concept
    using edge_iterator = std::nullptr_t;
    using edges_size_type = size_t;
};

// incidence graph concept

Graph::vertex_descriptor source(const Graph::edge_descriptor&, const Graph&) { return {}; }

Graph::vertex_descriptor target(const Graph::edge_descriptor&, const Graph&) { return {}; }

std::pair<Graph::out_edge_iterator, Graph::out_edge_iterator> out_edges(
    Graph::vertex_descriptor, const Graph&)
{
    return { {}, {} };
}

Graph::degree_size_type out_degree(Graph::vertex_descriptor, const Graph&) { return 0; }

// vertex list graph concept

std::pair<Graph::vertex_iterator, Graph::vertex_iterator> vertices(const Graph&)
{
    return { {}, {} };
}

Graph::vertices_size_type num_vertices(const Graph&) { return 0; }

// edge list graph concept

Graph::edges_size_type num_edges(const Graph&) { return 0; }

std::pair<Graph::edge_iterator, Graph::edge_iterator> edges(const Graph&) { return { {}, {} }; }

// adjacency matrix graph concept

std::pair<Graph::edge_descriptor, bool> edge(
    Graph::vertex_descriptor, Graph::vertex_descriptor, const Graph&)
{
    return { {}, {} };
}

// property maps for boykov_kolmogorov_max_flow

auto get(boost::edge_capacity_t, Graph&)
{
    return boost::make_constant_property<Graph::edge_descriptor>(0);
}

auto get(boost::edge_residual_capacity_t, Graph&)
{
    return boost::make_constant_property<Graph::edge_descriptor>(0);
}

auto get(boost::edge_reverse_t, const Graph&)
{
    return boost::make_constant_property<Graph::edge_descriptor>(Graph::edge_descriptor {});
}

auto get(boost::vertex_predecessor_t, Graph&)
{
    return boost::make_constant_property<Graph::vertex_descriptor>(Graph::edge_descriptor {});
}

auto get(boost::vertex_color_t, Graph&)
{
    return boost::make_constant_property<Graph::vertex_descriptor>(boost::default_color_type {});
}

auto get(boost::vertex_distance_t, Graph&)
{
    return boost::make_constant_property<Graph::vertex_descriptor>(0);
}

auto get(boost::vertex_index_t, Graph&)
{
    return boost::make_constant_property<Graph::vertex_descriptor>(0);
}

}

int main()
{
    my_graph::Graph G;

    boost::boykov_kolmogorov_max_flow(G, 0, 1);
}
```

[Live on compiler-explorer.](https://compiler-explorer.com/z/6sTej3axx)

sebrockm:
This was supposed to fix #309, but that one turned out to be not an issue.

However, there were no tests for the all defaults overload of `boykov_kolmogorov_max_flow` which I added. So maybe this PR is still worth being merged.

### DIFF
--- a/test/boykov_kolmogorov_max_flow_test.cpp
+++ b/test/boykov_kolmogorov_max_flow_test.cpp
@@ -211,11 +211,17 @@ long test_bundled_properties(int n_verts, int n_edges, std::size_t seed)
 long test_overloads(int n_verts, int n_edges, std::size_t seed)
 {
     typedef adjacency_list_traits< vecS, vecS, directedS > tTraits;
+    typedef property< vertex_index_t, long,
+        property< vertex_predecessor_t, tTraits::edge_descriptor,
+            property< vertex_color_t, default_color_type,
+                property< vertex_distance_t, long > > > >
+        tVertexProperty;
     typedef property< edge_capacity_t, long,
         property< edge_residual_capacity_t, long,
             property< edge_reverse_t, tTraits::edge_descriptor > > >
         tEdgeProperty;
-    typedef adjacency_list< vecS, vecS, directedS, no_property, tEdgeProperty >
+    typedef adjacency_list< vecS, vecS, directedS, tVertexProperty,
+        tEdgeProperty >
         tGraph;
 
     tGraph g;
@@ -240,7 +246,10 @@ long test_overloads(int n_verts, int n_edges, std::size_t seed)
             color_vec.begin(), get(vertex_index, g)),
         get(vertex_index, g), src, sink);
 
+    long flow_overload_3 = boykov_kolmogorov_max_flow(g, src, sink);
+
     BOOST_TEST(flow_overload_1 == flow_overload_2);
+    BOOST_TEST(flow_overload_2 == flow_overload_3);
     return flow_overload_1;
 }
 
--- a/test/boykov_kolmogorov_max_flow_test.cpp
+++ b/test/boykov_kolmogorov_max_flow_test.cpp
@@ -174,20 +174,117 @@ template < typename EdgeDescriptor > struct Link
     EdgeDescriptor edge_reverse;
 };
 
+typedef adjacency_list_traits< vecS, vecS, directedS > tTraits;
+typedef Node< tTraits::edge_descriptor > tVertex;
+typedef Link< tTraits::edge_descriptor > tEdge;
+typedef adjacency_list< vecS, vecS, directedS, tVertex, tEdge > tBundleGraph;
+
+namespace boost
+{
+template <> struct property_map< tBundleGraph, edge_capacity_t >
+{
+    typedef adj_list_edge_property_map< directed_tag, long, long&, size_t,
+        tEdge, long tEdge::* >
+        type;
+    typedef adj_list_edge_property_map< directed_tag, long, const long&, size_t,
+        const tEdge, long tEdge::* >
+        const_type;
+};
+
+template <> struct property_map< tBundleGraph, edge_residual_capacity_t >
+{
+    typedef adj_list_edge_property_map< directed_tag, long, long&, size_t,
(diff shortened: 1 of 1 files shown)


test/boykov_kolmogorov_max_flow_test.cpp
```diff
@@ -174,21 +174,118 @@ template < typename EdgeDescriptor > struct Link
...
+property_map< tBundleGraph, vertex_color_t >::type get(
+    vertex_color_t, tBundleGraph& g)
+{
+    return get(&tVertex::vertex_color, g);
+}
+property_map< tBundleGraph, vertex_distance_t >::type get(
+    vertex_distance_t, tBundleGraph& g)
+{
+    return get(&tVertex::vertex_distance, g);
+}
```
jeremy-murphy, line 276: Maybe I'm missing something, but do you actually need all the above code in order to test the all-defaults overload?

sebrockm: My understanding of this is that the all-defaults overload uses overloads of `get` (e.g `get(vertex_distance, g)`) to obtain all the needed properties from the graph. Also, it uses `property_map<Property, Graph>::type` to determine the return type of those `get` functions. So, if the graph doesn't provide all of these by itself, I have to provide them. E.g. `adjacency_list` with interior properties already provides all of these `get` overloads and `property_map` specializations (see the other added test below which works out of the box once all needed interior properties are provided). However, with bundled properties, or any other kind of custom graph, those overloads and specializations do not exist, so I have to provide them.
Of course, it's questionable whether one would really want to write all this boiler plate code just to be able to call a handier overload of `boykov_kolmogorov_max_flow`. But it should still work (and it does work).

Btw, in my original attempt I only provided the `get` overloads, but was not aware that I also needed to provide specializations for `property_map`. That's why I opened #309. The lack of tests for the all-defaults overload further strengthened my believe that it is a bug. During my attempt of fixing it, I figured out that I was missing the `property_map` specializations and that there is no bug.

Long story short: Yes, I think all of this boiler plate is necessary, but I'd be happy to be proven wrong.

jeremy-murphy: Thanks, @sebrockm , I appreciate the detailed explanation.

Issue #312, opened 2022-11-06
maybe-uninitialized warning in detail/adj_list_edge_iterator.hpp


mattpulver:
Getting this compiler warning w/ gcc 12.2.0 + Boost 1.80:
```
In member function ‘boost::detail::adj_list_edge_iterator<VertexIterator, OutEdgeIterator, Graph>::self& boost::detail::adj_list_edge_iterator<VertexIterator, OutEdgeIterator, Graph>::operator++() [with VertexIterator = boost::range_detail::integer_iterator<long unsigned int>; OutEdgeIterator = boost::detail::out_edge_iter<std::_List_iterator<boost::detail::stored_edge_property<long unsigned int, boost::property<boost::edge_weight_t, int> > >, long unsigned int, boost::detail::edge_desc_impl<boost::directed_tag, long unsigned int>, long int>; Graph = boost::adjacency_list<boost::listS, boost::vecS, boost::directedS, boost::no_property, boost::property<boost::edge_weight_t, int> >]’,
    inlined from ‘void boost::vec_adj_list_impl<Graph, Config, Base>::copy_impl(const boost::vec_adj_list_impl<Graph, Config, Base>&) [with Graph = boost::adjacency_list<boost::listS, boost::vecS, boost::directedS, boost::no_property, boost::property<boost::edge_weight_t, int> >; Config = boost::detail::adj_list_gen<boost::adjacency_list<boost::listS, boost::vecS, boost::directedS, boost::no_property, boost::property<boost::edge_weight_t, int> >, boost::vecS, boost::listS, boost::directedS, boost::no_property, boost::property<boost::edge_weight_t, int>, boost::no_property, boost::listS>::config; Base = boost::directed_graph_helper<boost::detail::adj_list_gen<boost::adjacency_list<boost::listS, boost::vecS, boost::directedS, boost::no_property, boost::property<boost::edge_weight_t, int> >, boost::vecS, boost::listS, boost::directedS, boost::no_property, boost::property<boost::edge_weight_t, int>, boost::no_property, boost::listS>::config>]’ at /usr/include/boost/graph/detail/adjacency_list.hpp:2193:63:
[...]

PR #313, opened 2022-11-14
Silence maybe-uninitialized warning in adj_list_edge_iterator::operator++()

mattpulver:
It's inconclusive whether or not this is a legitimate warning or a GCC 12 false positive. The warning given in #312 appears to only show when built in cmake Release mode.

### DIFF
--- a/include/boost/graph/detail/adjacency_list.hpp
+++ b/include/boost/graph/detail/adjacency_list.hpp
@@ -2189,8 +2189,12 @@ public:
         }
         // Copy the edges by adding each edge and copying its
         // property object.
+#ifdef BOOST_NO_CXX17_STRUCTURED_BINDINGS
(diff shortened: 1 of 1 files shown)


jeremy-murphy:
Now that I see exactly what you mean it does sound suspiciously like a false positive. 

I've found one recent discussion about false positives from GCC 12.1 (https://github.com/catchorg/Catch2/issues/2423), so I'm wondering maybe it's better to wait this one out for a while and see if it gets fixed in the 12.x series.

mattpulver:
Unfortunately I'm seeing this w/ `gcc 12.2.0`. But I'm fine to wait it out. I agree this is not the most desirable hack.

jeremy-murphy:
Let's draw a line and say that if it's not fixed by the time 13.1 comes out, then we implement our fix.

PR #315, opened 2022-11-26
Optionally specify a maximum circuit length in Hawick circuits algorithm

svengato:
The number of circuits grows rapidly for larger graphs, so this is a way to reduce the computation time and still obtain useful partial results.

### DIFF
--- a/include/boost/graph/hawick_circuits.hpp
+++ b/include/boost/graph/hawick_circuits.hpp
@@ -153,7 +153,7 @@ namespace hawick_circuits_detail
     public:
         hawick_circuits_from(Graph const& graph, Visitor& visitor,
             VertexIndexMap const& vim, Stack& stack, ClosedMatrix& closed,
-            VerticesSize n_vertices, unsigned int const max_length)
+            VerticesSize n_vertices, unsigned int max_length)
         : graph_(graph)
         , visitor_(visitor)
         , vim_(vim)
@@ -298,14 +298,14 @@ namespace hawick_circuits_detail
         Stack& stack_;
         ClosedMatrix& closed_;
         BlockedMap blocked_;
-        unsigned int const max_length_;
+        unsigned int max_length_;
     };
 
     template < typename GetAdjacentVertices, typename Graph, typename Visitor,
         typename VertexIndexMap >
     void call_hawick_circuits(Graph const& graph,
         Visitor /* by value */ visitor, VertexIndexMap const& vertex_index_map,
-        unsigned int const max_length)
+        unsigned int max_length)
     {
         typedef graph_traits< Graph > Traits;
         typedef typename Traits::vertex_descriptor Vertex;
@@ -348,7 +348,7 @@ namespace hawick_circuits_detail
     template < typename GetAdjacentVertices, typename Graph, typename Visitor >
     void call_hawick_circuits(
         Graph const& graph, BOOST_FWD_REF(Visitor) visitor,
-        unsigned int const max_length)
+        unsigned int max_length)
     {
         call_hawick_circuits< GetAdjacentVertices >(graph,
             boost::forward< Visitor >(visitor), get(vertex_index, graph),
@@ -360,7 +360,7 @@ namespace hawick_circuits_detail
 template < typename Graph, typename Visitor, typename VertexIndexMap >
 void hawick_circuits(BOOST_FWD_REF(Graph) graph, BOOST_FWD_REF(Visitor) visitor,
     BOOST_FWD_REF(VertexIndexMap) vertex_index_map,
-    unsigned int const max_length = 0)
+    unsigned int max_length = 0)
 {
     hawick_circuits_detail::call_hawick_circuits<
         hawick_circuits_detail::get_all_adjacent_vertices >(
@@ -370,7 +370,7 @@ void hawick_circuits(BOOST_FWD_REF(Graph) graph, BOOST_FWD_REF(Visitor) visitor,
 
 template < typename Graph, typename Visitor >
 void hawick_circuits(BOOST_FWD_REF(Graph) graph, BOOST_FWD_REF(Visitor) visitor,
-    unsigned int const max_length = 0)
+    unsigned int max_length = 0)
 {
     hawick_circuits_detail::call_hawick_circuits<
         hawick_circuits_detail::get_all_adjacent_vertices >(
@@ -386,7 +386,7 @@ template < typename Graph, typename Visitor, typename VertexIndexMap >
 void hawick_unique_circuits(BOOST_FWD_REF(Graph) graph,
     BOOST_FWD_REF(Visitor) visitor,
     BOOST_FWD_REF(VertexIndexMap) vertex_index_map,
-    unsigned int const max_length = 0)
+    unsigned int max_length = 0)
 {
     hawick_circuits_detail::call_hawick_circuits<
         hawick_circuits_detail::get_unique_adjacent_vertices >(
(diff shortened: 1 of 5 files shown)


jeremy-murphy:
Looks good but needs at least one unit test to demonstrate that it works, i.e. showing that a non-zero `maxLength` will return a suboptimal answer.

include/boost/graph/hawick_circuits.hpp
```diff
@@ -153,13 +153,14 @@ namespace hawick_circuits_detail
     public:
         hawick_circuits_from(Graph const& graph, Visitor& visitor,
             VertexIndexMap const& vim, Stack& stack, ClosedMatrix& closed,
-            VerticesSize n_vertices)
+            VerticesSize n_vertices, unsigned int const maxLength)
```
jeremy-murphy, line 156: Don't `const` value parameters.
```suggestion
            VerticesSize n_vertices, unsigned int maxLength)
```

include/boost/graph/hawick_circuits.hpp
```diff
@@ -333,30 +347,35 @@ namespace hawick_circuits_detail
...
         hawick_circuits_detail::get_all_adjacent_vertices >(
         boost::forward< Graph >(graph), boost::forward< Visitor >(visitor),
-        boost::forward< VertexIndexMap >(vertex_index_map));
+        boost::forward< VertexIndexMap >(vertex_index_map), maxLength);
 }
 
 template < typename Graph, typename Visitor >
-void hawick_circuits(BOOST_FWD_REF(Graph) graph, BOOST_FWD_REF(Visitor) visitor)
+void hawick_circuits(BOOST_FWD_REF(Graph) graph, BOOST_FWD_REF(Visitor) visitor,
+    unsigned int const maxLength = 0)
```
jeremy-murphy, line 373:
```suggestion
    unsigned int maxLength = 0)
```

include/boost/graph/hawick_circuits.hpp
```diff
@@ -366,21 +385,24 @@ void hawick_circuits(BOOST_FWD_REF(Graph) graph, BOOST_FWD_REF(Visitor) visitor)
 template < typename Graph, typename Visitor, typename VertexIndexMap >
 void hawick_unique_circuits(BOOST_FWD_REF(Graph) graph,
     BOOST_FWD_REF(Visitor) visitor,
-    BOOST_FWD_REF(VertexIndexMap) vertex_index_map)
+    BOOST_FWD_REF(VertexIndexMap) vertex_index_map,
+    unsigned int const maxLength = 0)
```
jeremy-murphy, line 389:
```suggestion
    unsigned int maxLength = 0)
```

include/boost/graph/hawick_circuits.hpp
```diff
@@ -366,21 +385,24 @@ void hawick_circuits(BOOST_FWD_REF(Graph) graph, BOOST_FWD_REF(Visitor) visitor)
...
         boost::forward< Graph >(graph), boost::forward< Visitor >(visitor),
-        boost::forward< VertexIndexMap >(vertex_index_map));
+        boost::forward< VertexIndexMap >(vertex_index_map), maxLength);
 }
 
 template < typename Graph, typename Visitor >
 void hawick_unique_circuits(
-    BOOST_FWD_REF(Graph) graph, BOOST_FWD_REF(Visitor) visitor)
+    BOOST_FWD_REF(Graph) graph, BOOST_FWD_REF(Visitor) visitor,
+    unsigned int const maxLength = 0)
```
jeremy-murphy, line 400:
```suggestion
    unsigned int maxLength = 0)
```

include/boost/graph/hawick_circuits.hpp
svengato: Thanks, I am removing the `const` from the `max_length` function arguments, plus some minor cleanup (like `max_length` instead of `maxLength`).

One question: what about the data member `max_length_`? Should that remain `const`?

svengato:
Unit tests work. Just waiting on the `const` data member question -- one thing I read suggests data members should not be `const` because then the data structure is not copy-assignable.

svengato:
> Just waiting on the `const` data member question --

I decided to remove that last `const` and push the commits, that will give you more to work with. Let me know if any of the `const` need to be restored.

jeremy-murphy:
> Unit tests work. Just waiting on the `const` data member question -- one thing I read suggests data members should not be `const` because then the data structure is not copy-assignable.

Correct, but I think that class also has member references that will also prevent copying, so I didn't both to comment on it.

jeremy-murphy:
> Unit tests work. Just waiting on the `const` data member question -- one thing I read suggests data members should not be `const` because then the data structure is not copy-assignable.

Unfortunately, it looks like that `cycle_test` function is not really _testing_ the functionality, just reporting something to the console?
What we need is a test that will produce a different result for `hawick_(unique_)circuits` depending on whether it was called with a custom `max_length` or not and fail if the wrong result is returned. Maybe I'm missing something, but I don't see that here.

svengato:
But it is reporting the correct result, the number of circuits with length up to 4.

svengato:
The current tests in `test/cycle_test.hpp` and `test/hawick_circuits.cpp` do not return a pass/fail value either. Are you looking for something like this?
```
// undirected, not unique, no maximum length
BOOST_TEST(vis.cycles == 30);
// directed, not unique, no maximum length
BOOST_TEST(vis.cycles == 31); 

// undirected, unique, no maximum length
BOOST_TEST(vis.cycles == 27);
// directed, unique, no maximum length
BOOST_TEST(vis.cycles == 24);

// undirected, not unique, maximum length 4
BOOST_TEST(vis.cycles == 28);
// directed, not unique, maximum length 4
BOOST_TEST(vis.cycles == 13);

// undirected, unique, maximum length 4
BOOST_TEST(vis.cycles == 25);
// directed, unique, maximum length 4
BOOST_TEST(vis.cycles == 10);
```

jeremy-murphy:
> The current tests in `test/cycle_test.hpp` and `test/hawick_circuits.cpp` do not return a pass/fail value either. Are you looking for something like this?
> 
> ```
> // undirected, not unique, no maximum length
> BOOST_TEST(vis.cycles == 30);
> // directed, not unique, no maximum length
> BOOST_TEST(vis.cycles == 31); 
> 
> // undirected, unique, no maximum length
> BOOST_TEST(vis.cycles == 27);
> // directed, unique, no maximum length
> BOOST_TEST(vis.cycles == 24);
> 
> // undirected, not unique, maximum length 4
> BOOST_TEST(vis.cycles == 28);
> // directed, not unique, maximum length 4
> BOOST_TEST(vis.cycles == 13);
> 
> // undirected, unique, maximum length 4
> BOOST_TEST(vis.cycles == 25);
> // directed, unique, maximum length 4
> BOOST_TEST(vis.cycles == 10);
> ```

Yes. And we should be able to demonstrate that the result changes appropriately as the value for `max_length` changes.

svengato:
The current tests are for no `max_length` and `max_length` = 4. Would you prefer that it test all `max_length` from 0 to 10 (length of the test graph's longest directed circuit)? That would be easy to implement.

jeremy-murphy:
> The current tests are for no `max_length` and `max_length` = 4. Would you prefer that it test all `max_length` from 0 to 10 (length of the test graph's longest directed circuit)? That would be easy to implement.

Hey sorry for the really delayed reply. Yes, I think testing a variety of values is better, and 0 to 10 sounds good without being excessive.

jeremy-murphy:
Awesome, thanks for the great idea and contribution!

svengato:
You are welcome!

PR #321, opened 2022-12-28
Fix hyperlink for disjoint sets in "Challenge and To-Do Items"

derek-mcblane:
The link text was messed up because the tag isn't closed.

### DIFF
--- a/doc/challenge.html
+++ b/doc/challenge.html
@@ -37,7 +37,7 @@ review. The pending items are:</li>
      <tt>mutable_queue.hpp</tt>, <tt>fibonacci_heap.hpp</tt>.
       Somehow merge implementation with Dietmer's heaps and queues.</li>
 
-   <li><tt>disjoint_sets</tt> (see <a href="disjoint_sets.html">)</li>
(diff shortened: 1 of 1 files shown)


doc/challenge.html
```diff
@@ -37,7 +37,7 @@ <h2>Boost Graph Library Challenge and To-Do Items</h2>
      <tt>mutable_queue.hpp</tt>, <tt>fibonacci_heap.hpp</tt>.
       Somehow merge implementation with Dietmer's heaps and queues.</li>
 
-   <li><tt>disjoint_sets</tt> (see <a href="disjoint_sets.html">)</li>
+   <li><tt>disjoint_sets</tt> (see <a href="disjoint_sets.html">Disjoint Sets</a>)</li>
```
jeremy-murphy, line 40: I'm fairly naive about HTML, so can you explain to me why there needs to be a `</a>`? I thought the opening `<a` was closed by the 'naked' `>`?

jeremy-murphy:
Thanks for fixing it!

PR #325, opened 2023-02-10
Add missing include on boost/config.hpp

ams2990:
exception.hpp uses BOOST_SYMBOL_VISIBLE, which is defined in boost/config.hpp. Without this, compiling just exception.hpp is broken.


jeremy-murphy:
Thanks!

PR #331, opened 2023-03-19
Don't look for disjoint-sets representatives in link

### LINKED ISSUE #327: Redundant search for representative in disjoint sets
We have a general function to join 2 sets
https://github.com/boostorg/graph/blob/00028767d804fe3da20996afedf06f9eecfb5975/include/boost/pending/disjoint_sets.hpp#L78-L82
This function finds the representative for each set, then calls the more specialized
https://github.com/boostorg/graph/blob/00028767d804fe3da20996afedf06f9eecfb5975/include/boost/pending/disjoint_sets.hpp#L72-L76
which requires that we pass the representative of each set. It then calls the worker
https://github.com/boostorg/graph/blob/00028767d804fe3da20996afedf06f9eecfb5975/include/boost/pending/detail/disjoint_sets.hpp#L55-L59
which starts by looking for the representative of each set... (if anything, this could be `assert(i == comp_rep(p, i));`)

It isn't horrible, the search of the representative only takes one query in parent to check that it is indeed already good, it is probably already in cache and the branch predictor probably learns that this comparison is always true. Still, this defeats the purpose of having 2 different functions.

Note that fixing this has the potential to break code for users who mistakenly used link instead of union_sets, and didn't notice since they are currently equivalent.

mglisse:
The arguments are already required to be the representative of their class. If we do not know the representatives, we are supposed to call union_set instead, which first finds them.
Fix #327. In my application (very similar to minimum spanning tree), this saves about 5% of the total running time, which is small but not negligible. That was with full path compression, I think the gains are larger with path halving, where a second consecutive call is non-trivial.
In example/incremental_components.cpp, there is a DS with vertex type size_t, but link is called on plain int, which can cause some conversion warnings. After the patch, in debug mode, it may become a warning about comparing the 2 types. Depending on your compiler and options, it may thus give a new warning where there was none, but in my opinion that's independent from this PR.

### DIFF
--- a/include/boost/pending/detail/disjoint_sets.hpp
+++ b/include/boost/pending/detail/disjoint_sets.hpp
@@ -6,6 +6,8 @@
 #ifndef BOOST_DETAIL_DISJOINT_SETS_HPP
 #define BOOST_DETAIL_DISJOINT_SETS_HPP
 
+#include <cassert>
+
 namespace boost
 {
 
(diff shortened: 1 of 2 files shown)


jeremy-murphy:
Thanks for improving this!

PR #332, opened 2023-03-27
Add missing break to switch

ams2990:
Clang warns about this with `-Wimplicit-fallthrough`.


jeremy-murphy:
Why not just add a `break` statement? Seems cleaner to me. After all, there is nothing in the `default` case that requires falling through into it.

ams2990:
Done.

jeremy-murphy:
Thanks for that!

Issue #297, opened 2022-07-11
`maximum_adjacency_search`: dead code, broken `assignments` parameter, missing contraction support


sebrockm:
There are several things wrong with `maximum_adjacency_search`:

- The pseudo code on https://www.boost.org/doc/libs/1_79_0/libs/graph/doc/maximum_adjacency_search.html does not mention any weights. Yet, `weights` is a required parameter. It's unclear what it is doing.
- In [L127](https://github.com/boostorg/graph/blob/430f61f6962614bb33fe7be3b054229953803ced/include/boost/graph/maximum_adjacency_search.hpp#L127), there is a `set` defined, but it is never filled with any values. In [L181](https://github.com/boostorg/graph/blob/430f61f6962614bb33fe7be3b054229953803ced/include/boost/graph/maximum_adjacency_search.hpp#L181) there is a loop iterating over this always empty `set`. This is dead code.
- It looks like the `assignments` property map is a dead parameter. Looking at the implementation, in [L131](https://github.com/boostorg/graph/blob/430f61f6962614bb33fe7be3b054229953803ced/include/boost/graph/maximum_adjacency_search.hpp#L131) we see the one and only `put` operation performed on it. It overrides all possibly provided user input and assigns every vertex to itself. That makes all following `get` operations obsolete (every `get(assignments, x)` is equivalent to just `x`).
- [L158](https://github.com/boostorg/graph/blob/430f61f6962614bb33fe7be3b054229953803ced/include/boost/graph/maximum_adjacency_search.hpp#L158) is a noop

### Background:
This algorithm was introduced in https://github.com/boostorg/graph/commit/393c072c186f6711d1537225dc116ea1528140ac. There, it was extracted from `stoer_wagner_min_cut` which caused #286.

jeremy-murphy:
Interesting. Thanks for looking into it! You've certainly identified some serious issues with that implementation.

The next big question would be, do the existing unit tests cover everything? If they do, then someone can be free to refactor the hell out of that implementation. But I can only imagine that it is going to require some additional tests to cover the obviously wrong behaviour on `assignments`, etc.

The existing unit tests are moderately complicated; if someone wants to tackle this, I'd recommend adding some ultra-simple ones too.

daank94:
[...]

PR #333, opened 2023-04-23
Maximum Adjacency Search: refactor existing tests + create new tests

daankolthof:
See https://github.com/boostorg/graph/issues/297

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,53 +56,6 @@ jobs:
       - name: Test
         run: ../../../b2 toolset=$TOOLSET cxxstd=03,11,14,17,2a
         working-directory: ../boost-root/libs/graph/test
-  ubuntu-bionic:
-    runs-on: ubuntu-18.04
-    strategy:
-      fail-fast: false
-      matrix:
-        compiler: [ g++-7, g++-8, clang++-7, clang++-8 ]
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          fetch-depth: '0'
-      - uses: mstachniuk/ci-skip@v1
-        with:
-          commit-filter: '[skip ci];[ci skip];[CI SKIP];[SKIP CI];***CI SKIP***;***SKIP CI***;[windows];[Windows];[WINDOWS];[apple];[Apple];[APPLE]'
-          commit-filter-separator: ';'
-          fail-fast: true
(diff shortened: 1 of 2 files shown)


daankolthof:
Blocked on https://github.com/boostorg/graph/issues/334.

test/mas_test.cpp
```diff
@@ -44,54 +46,45 @@ struct edge_t
...
+        return *vertex_weights_when_visited_;
+    }
 
 private:
-    const KeyedUpdatablePriorityQueue& m_pq;
-    boost::shared_ptr< vertex_descriptor > m_curr, m_prev;
-    boost::shared_ptr< weight_type > m_reach_weight;
+    const KeyedUpdatablePriorityQueue& m_pq_;
+    boost::shared_ptr< std::vector< vertex_descriptor > > vertex_visit_order_;
+    boost::shared_ptr< std::vector< weight_type > > vertex_weights_when_visited_;
```
jeremy-murphy, line 87: Do these have to be `shared_ptr`? I don't have time to look at the full code, but I couldn't see a reason for it here in the PR code.

daankolthof: These are `shared_ptr`'s because the visitor object is passed by value during the call to `maximum_adjacency_search`. Apparently this passing by value is 'normal' as I'm also seeing it in BFS.

In DFS however, it's passing by reference, which makes more sense to me. This might be a separate future work itme.

daankolthof:
> Looks like the build failed on something unrelated to your code. Can you trigger a rebuild?

Only by pushing a new commit, I don't have any access to re-trigger the failed build. Do you have access?

jeremy-murphy:
> > Looks like the build failed on something unrelated to your code. Can you trigger a rebuild?
> 
> Only by pushing a new commit, I don't have any access to re-trigger the failed build. Do you have access?

I never figured out if it was possible. But now that I've merged the fix for Bionic, you can merge that change in and push it.

test/mas_test.cpp
jeremy-murphy: Ah, I see. Yeah, I think those algorithms should probably take the visitor by value _and_ return it by value.

daankolthof:
I've switched some branched around on my fork, new PR is at https://github.com/boostorg/graph/pull/337.

PR #336, opened 2023-04-25
Ubuntu Bionic is deprecated and GitHub will no longer host CI workers running on it

daankolthof:
GitHub blog: https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

Also fixes:
- Deprecation of boost/progress.hpp

Hopefully after this PR is merged the CI builds actually succeed for the develop branch. After that, new features can be implemented.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,53 +56,6 @@ jobs:
       - name: Test
         run: ../../../b2 toolset=$TOOLSET cxxstd=03,11,14,17,2a
         working-directory: ../boost-root/libs/graph/test
-  ubuntu-bionic:
-    runs-on: ubuntu-18.04
-    strategy:
-      fail-fast: false
-      matrix:
-        compiler: [ g++-7, g++-8, clang++-7, clang++-8 ]
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          fetch-depth: '0'
-      - uses: mstachniuk/ci-skip@v1
-        with:
-          commit-filter: '[skip ci];[ci skip];[CI SKIP];[SKIP CI];***CI SKIP***;***SKIP CI***;[windows];[Windows];[WINDOWS];[apple];[Apple];[APPLE]'
-          commit-filter-separator: ';'
-          fail-fast: true
-      - name: Set TOOLSET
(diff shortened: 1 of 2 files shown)


daankolthof:
@jeremy-murphy I would like to have this pre-approved as it's blocking everything else and thus needs to be merged as soon as possible.

jeremy-murphy:
Looks good, but it's late at night for me so I'll merge it in the morning after I have a complete check.

jeremy-murphy:
Awesome, thanks for making those changes!

daankolthof:
Thanks for merging this @jeremy-murphy, maybe we add the latest versions of the GitHub Action Runners? See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.
This obviously depends on Boost's strategy for integration testing, which I don't know much about.

jeremy-murphy:
> Thanks for merging this @jeremy-murphy, maybe we add the latest versions of the GitHub Action Runners? See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources. This obviously depends on Boost's strategy for integration testing, which I don't know much about.

The Boost organization provides resources, but each library has their own strategy.

I support adding the latest Ubuntu image as a runner. Let's not use the `xxx-latest` label, though, as that causes unwanted surprises when they change it to point to a new image.  :)

daankolthof:
> > Thanks for merging this @jeremy-murphy, maybe we add the latest versions of the GitHub Action Runners? See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources. This obviously depends on Boost's strategy for integration testing, which I don't know much about.
> 
> The Boost organization provides resources, but each library has their own strategy.
> 
> I support adding the latest Ubuntu image as a runner. Let's not use the `xxx-latest` label, though, as that causes unwanted surprises when they change it to point to a new image. :)

Sounds good! If you could create an Issue, I can pick it up sometime in May.

PR #339, opened 2023-05-08
Make comments match actual values of the random graph's number of ver…

svengato:
…tices and probability of edge connection

### DIFF
--- a/test/bron_kerbosch_all_cliques.cpp
+++ b/test/bron_kerbosch_all_cliques.cpp
@@ -51,7 +51,7 @@ template < typename Graph > void test()
 {
     typedef erdos_renyi_iterator< boost::minstd_rand, Graph > er;
 
-    // Generate random graphs with 15 vertices and 15% probability
+    // Generate random graph with 20 vertices and 10% probability
     // of edge connection.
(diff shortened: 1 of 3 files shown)


svengato:
I noticed the mismatches while working on pull request #315. The changes are only to comments and should not require testing.

jeremy-murphy:
Even better might be to remove the numbers from the comments altogether?

E.g.:
```
    // Generate random graph with N vertices and probability P
    // of edge connection.
    static const size_t N = 20;
    static const double P = 0.1;
```
or
```
    // Generate random graph
    static const size_t N = 20; // Number of vertices
    static const double P = 0.1; // Probability of edge between two vertices
```

jeremy-murphy:
Btw, there is a way to tell the CI not to run, which would be appropriate for these changes, by including some magic text in the commit message. I don't remember exactly what it is... instructions are probably on GitHub somewhere.

jeremy-murphy:
Thanks again!

svengato:
> Btw, there is a way to tell the CI not to run, [...]

Added to my **Things to learn** list. Thanks -

PR #341, opened 2023-07-07
Compress boost::property

### LINKED ISSUE #142: compress boost::property
(adapted from https://svn.boost.org/trac10/ticket/9826)

Looking at property_map, I noticed that boost::property is wasting space. property<tag,double> will typically have size 16 because of the no_property member. list_edge<Vertex,no_property> will also be too long for the same reason. A proper use of the empty base optimization (possibly through tuple or compressed_pair) should help with this. Partial specializations for no_property may be a simpler alternative. If you are lazy, [[no_unique_address]] would already be nice.

Graphs can grow large, and it seems quite important to avoid wasting memory.

This appears to be a regression caused by:

> commit 616b9e71348b430af89a2405b5b8463557e63dec
> Author: Jeremiah Willcock <jewillco@osl.iu.edu>
> Date:   Sat Jun 30 20:00:41 2012 +0000

Before that commit, property was implemented with derivation (which is still what the documentation shows...), and the empty base optimization avoided wasting so much space.

mglisse:
Fix #142.
It isn't the nicest fix, but it is the smallest one I can think of and someone is free to do better later. Specializing `property<*,*,no_property>` without m_base would require changing the lookup code that expects this member. Grouping the 2 members in a tuple would also require (syntactic) changes in the rest of the file.

I am not adding a testcase like
```c++
#include <boost/static_assert.hpp>
#include <boost/pending/property.hpp>

struct Tag {};
struct Reference { double d; };
typedef boost::property<Tag, Reference> Prop;

int main()
{
  BOOST_STATIC_ASSERT(sizeof(Reference) == sizeof(Prop));
}
```
because I don't think Boost.Config provides an easy way to detect platforms where BOOST_ATTRIBUTE_NO_UNIQUE_ADDRESS does something.


jeremy-murphy:
We don't have any C++23 builds in our CI at the moment, so you'll have to wait for that to be added via a separate PR.

mglisse:
> We don't have any C++23 builds in our CI at the moment, so you'll have to wait for that to be added via a separate PR.

BOOST_ATTRIBUTE_NO_UNIQUE_ADDRESS expands to `[[no_unique_address]]` for both gcc and clang (except possibly on windows?) even with `-std=c++11`, and the attribute is not ignored.

jeremy-murphy:
OK, let's give it a go, thanks!

PR #348, opened 2023-09-04
Remove GCC 4.x, Clang 3.5 and C++03

jeremy-murphy:
Remove testing of '98 and '03 language standards now that they are officially deprecated in Boost.

Also removed the oldest GCC (4.x) and Clang (3.5) versions, somewhat arbitrarily. At some point I'll try to find a consensus on how many years we should remain backward compatible with compilers for.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -14,17 +14,6 @@ windowsglobalimage="cppalliance/dronevs2019"
 
 def main(ctx):
   return [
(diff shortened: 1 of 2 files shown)


jeremy-murphy:
@jzmaddock , I'm going to merge this now because it's kind of an urgent fix, but if you think I made a mistake removing the old GCC and Clang, let me know.

jeremy-murphy:
> Should probably be updated for more recent releases too....

Yeah, I'll do that next. Just needed to get the build green this time.

PR #349, opened 2023-09-07
Isomorphism: Remove invariant contiguous range requirement

jan-grimo:
Completes isomorphism invariant improvements outlined in #203 

Invariant contiguous range requirement removal
- Vertex invariants for use in isomorphism algorithm must no longer have low upper bounds due to a hidden allocation linear in the maximum encountered vertex invariant.
- Vertex invariants must no longer be convertible to `size_t`, but can be any comparable and hashable types
- Build `unordered_map`-backed invariant multiplicity map efficiently from sorted vertex invariants

### DIFF
--- a/include/boost/graph/isomorphism.hpp
+++ b/include/boost/graph/isomorphism.hpp
@@ -165,29 +165,26 @@ namespace detail
         template<typename ForwardIterator>
         multiplicity_map multiplicities(ForwardIterator first, const ForwardIterator last) 
         {
+            typedef typename multiplicity_map::iterator invar_map_iter;
+
             assert(std::is_sorted(first, last));
             multiplicity_map invar_multiplicity;
 
-            typedef typename std::vector< invariant_t >::const_iterator invar_iter;
-            typedef typename multiplicity_map::iterator invar_map_iter;
-
             if(first == last) 
                 return invar_multiplicity;
-            
-            invar_iter it = std::move(first);
 
-            invariant_t invar = *it;
+            invariant_t invar = *first;
             invar_map_iter inserted = invar_multiplicity.emplace(invar, 1).first;
-            ++it;
-            for(; it != last; ++it)
+            ++first;
+            for(; first != last; ++first)
             {
-                if(*it == invar)
+                if(*first == invar)
                 {
                     inserted->second += 1;
                 }
                 else
                 {
-                    invar = *it;
+                    invar = *first;
                     inserted = invar_multiplicity.emplace(invar, 1).first;
                 }
             }
--- a/include/boost/graph/isomorphism.hpp
+++ b/include/boost/graph/isomorphism.hpp
@@ -30,6 +30,22 @@ namespace boost
 namespace detail
 {
 
+    template < typename T >
+    struct HashableConcept {
+        BOOST_CONCEPT_USAGE(HashableConcept)
+        {
+            hash<T> hasher;
+            typedef typename hash<T>::result_type hash_result;
+            hash_result val = hasher(t);
+            boost::ignore_unused_variable_warning(val);
+        }
+
+        T t;
+    };
+
+    template < typename Invariant >
+    struct InvariantConcept : LessThanComparable<Invariant>, EqualityComparable<Invariant>, HashableConcept<Invariant> {};
+
     template < typename Graph1, typename Graph2, typename IsoMapping,
         typename Invariant1, typename Invariant2, typename IndexMap1,
         typename IndexMap2 >
@@ -39,8 +55,8 @@ namespace detail
         typedef typename graph_traits< Graph2 >::vertex_descriptor vertex2_t;
         typedef typename graph_traits< Graph1 >::edge_descriptor edge1_t;
         typedef typename graph_traits< Graph1 >::vertices_size_type size_type;
-        typedef typename Invariant1::result_type invar1_value;
-        typedef typename Invariant2::result_type invar2_value;
+        typedef typename Invariant1::result_type invariant_t;
+        typedef unordered_map< invariant_t, size_type > multiplicity_map;
 
         const Graph1& G1;
         const Graph2& G2;
@@ -81,17 +97,17 @@ namespace detail
         friend struct compare_multiplicity;
         struct compare_multiplicity
         {
-            compare_multiplicity(Invariant1 invariant1, size_type* multiplicity)
+            compare_multiplicity(Invariant1 invariant1, const multiplicity_map& multiplicity)
             : invariant1(invariant1), multiplicity(multiplicity)
             {
             }
             bool operator()(const vertex1_t& x, const vertex1_t& y) const
             {
-                return multiplicity[invariant1(x)]
-                    < multiplicity[invariant1(y)];
+                return multiplicity.at(invariant1(x))
+                    < multiplicity.at(invariant1(y));
             }
             Invariant1 invariant1;
-            size_type* multiplicity;
+            const multiplicity_map& multiplicity;
         };
 
         struct record_dfs_order : default_dfs_visitor
@@ -158,46 +174,65 @@ namespace detail
             );
         }
 
+        // Generates map of invariant multiplicity from sorted invariants
+        multiplicity_map multiplicities(const std::vector< invariant_t >& invariants) {
+          // Assumes invariants are sorted
+          multiplicity_map invar_multiplicity;
+
+          typedef typename std::vector< invariant_t >::const_iterator invar_iter;
+          typedef typename multiplicity_map::iterator invar_map_iter;
+          invar_iter it = invariants.begin();
+          const invar_iter end = invariants.end();
+
+          if(it == end) {
+            return invar_multiplicity;
+          }
+
+          invariant_t invar = *it;
+          invar_map_iter inserted = invar_multiplicity.emplace(invar, 1).first;
+          ++it;
+          for(; it != end; ++it)
+          {
+              if(*it == invar)
+              {
+                  inserted->second += 1;
+              }
+              else
+              {
+                  invar = *it;
+                  inserted = invar_multiplicity.emplace(invar, 1).first;
+              }
+          }
+
+          return invar_multiplicity;
+        }
+
         bool test_isomorphism()
         {
             // reset isomapping
             BGL_FORALL_VERTICES_T(v, G1, Graph1)
             f[v] = graph_traits< Graph2 >::null_vertex();
 
-            std::size_t max_invariant = 0;
-            {
-                std::vector< invar1_value > invar1_array;
-                invar1_array.reserve(num_vertices(G1));
-                BGL_FORALL_VERTICES_T(v, G1, Graph1)
-                  invar1_array.push_back(invariant1(v));
-                sort(invar1_array);
+            // Calculate all invariants of G1 and G2, sort and compare
+            std::vector< invariant_t > invar1_array;
+            invar1_array.reserve(num_vertices(G1));
+            BGL_FORALL_VERTICES_T(v, G1, Graph1)
+            invar1_array.push_back(invariant1(v));
+            sort(invar1_array);
 
-                std::vector< invar2_value > invar2_array;
-                invar2_array.reserve(num_vertices(G2));
-                BGL_FORALL_VERTICES_T(v, G2, Graph2)
-                  invar2_array.push_back(invariant2(v));
-                sort(invar2_array);
-                if (!equal(invar1_array, invar2_array))
-                    return false;
+            std::vector< invariant_t > invar2_array;
+            invar2_array.reserve(num_vertices(G2));
+            BGL_FORALL_VERTICES_T(v, G2, Graph2)
+            invar2_array.push_back(invariant2(v));
+            sort(invar2_array);
+            if (!equal(invar1_array, invar2_array))
+                return false;
 
-                // Empty graphs case is covered before test_isomorphism is
-                // called, so the invar?_arrays cannot be empty, so back() is
-                // safe. Also the two invariant arrays are equal:
-                max_invariant = invar1_array.back();
-                assert(max_invariant != std::numeric_limits<std::size_t>::max());
-                max_invariant += 1;
-            }
-
+            // Sort vertices by the multiplicity of their invariants
             std::vector< vertex1_t > V_mult;
             BGL_FORALL_VERTICES_T(v, G1, Graph1)
             V_mult.push_back(v);
-            {
-                std::vector< size_type > multiplicity(max_invariant, 0);
-                BGL_FORALL_VERTICES_T(v, G1, Graph1)
-                ++multiplicity.at(invariant1(v));
-                sort(
-                    V_mult, compare_multiplicity(invariant1, &multiplicity[0]));
-            }
+            sort(V_mult, compare_multiplicity(invariant1, multiplicities(invar1_array)));
 
             std::vector< default_color_type > color_vec(num_vertices(G1));
             safe_iterator_property_map<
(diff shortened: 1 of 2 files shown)


jeremy-murphy:
Thanks for finishing this off!

I might not be able to merge this until I fix the modernization issues that are causing our CI builds to fail.

jargonzombies:
Oh, I thought that was me! No worries, I'm not in a hurry.

include/boost/graph/isomorphism.hpp
```diff
@@ -30,6 +30,22 @@ namespace boost
...
+    struct HashableConcept {
+        BOOST_CONCEPT_USAGE(HashableConcept)
+        {
+            hash<T> hasher;
+            typedef typename hash<T>::result_type hash_result;
+            hash_result val = hasher(t);
+            boost::ignore_unused_variable_warning(val);
+        }
+
+        T t;
```
jeremy-murphy, line 43: Maybe I'm being too pedantic here, but declaring `T t;` here requires that `T` is `DefaultConstructible`, but is that necessary for the `Hashable` concept?

include/boost/graph/isomorphism.hpp
```diff
@@ -81,17 +97,17 @@ namespace detail
...
             : invariant1(invariant1), multiplicity(multiplicity)
             {
             }
             bool operator()(const vertex1_t& x, const vertex1_t& y) const
             {
-                return multiplicity[invariant1(x)]
-                    < multiplicity[invariant1(y)];
+                return multiplicity.at(invariant1(x))
+                    < multiplicity.at(invariant1(y));
             }
```
jeremy-murphy, line 108: It's a bug if this function is called with invalid `x` or `y`, right? In which case, we should `assert` that they meet the relevant preconditions and ... I was going to say use member `operator[]`, but that's a bit misleading... so I guess it's member `find` and just assume that the iterator returned is not the end.
But basically, let's not use `at` if we're sure that it should be there.

include/boost/graph/isomorphism.hpp
```diff
@@ -81,17 +97,17 @@ namespace detail
...
             Invariant1 invariant1;
-            size_type* multiplicity;
+            const multiplicity_map& multiplicity;
```
jeremy-murphy, line 110: References break regular copy semantics, so just use a pointer.

include/boost/graph/isomorphism.hpp
```diff
@@ -158,46 +174,65 @@ namespace detail
             );
         }
 
+        // Generates map of invariant multiplicity from sorted invariants
+        multiplicity_map multiplicities(const std::vector< invariant_t >& invariants) {
```
jeremy-murphy, line 178: Implement this as
```suggestion
        template <typename ForwardIterator>
        multiplicity_map multiplicities(ForwardIterator first, ForwardIterator last) {
```

include/boost/graph/isomorphism.hpp
```diff
@@ -158,46 +174,65 @@ namespace detail
...
+          // Assumes invariants are sorted
```
jeremy-murphy, line 179: Don't just assume it, `assert` it!

include/boost/graph/isomorphism.hpp
```diff
@@ -158,46 +174,65 @@ namespace detail
...
+                  inserted->second += 1;
+              }
+              else
+              {
+                  invar = *it;
+                  inserted = invar_multiplicity.emplace(invar, 1).first;
+              }
+          }
+
+          return invar_multiplicity;
```
jeremy-murphy, line 207: So wait, this looks like it can be simplified to just call `invar_multiplicity[invar]++`, because if `invar` doesn't already exist in the map then `operator[]` creates it. Which means this function boils down to `std::for_each` with a pretty simple lambda like `[](auto invariant){ invariant_map[invariant]++; }`. 
Now, since Boost officially deprecated C++03, I think you can feel free to use C++14 if that simplifies the implementation, but don't do any mass conversion from C++03 to C++14.

jeremy-murphy:
Btw, we generally put the `{` on a new line after `if`, etc, so please keep to that format.

include/boost/graph/isomorphism.hpp
jargonzombies: It's not exactly the same! So the sorted list of invariants is laid out like e.g. `2, 2, 2, 7, 7, 9, 13` etc., right? The proposed version above doesn't repeatedly find the key position in the map for equal successive invariants, while a `for_each` version does. I assume it makes more of a performance difference the more repeated values there are. The default vertex invariant is the vertex degree, of which there are probably many equal values in large graphs. I haven't benchmarked anything, but I'm perfectly willing to.

include/boost/graph/isomorphism.hpp
jargonzombies: No, that's a good catch, it's unnecessary. I have to admit I found this exact concept in some other area of boost and didn't look at it closely enough.

jargonzombies: I can't use `declval` in this context though to avoid implying a particular construction method. I can avoid it by refactoring the whole invariant concept checking with C++11 features and `boost/type_traits` like so:

```c++
        template < typename T >
        using TestHashable = decltype(hash<T>()(std::declval<T>()));

        template < typename T >
        struct Hashable: std::integral_constant<bool, boost::is_detected_v<TestHashable, T>> {};

        template < typename T >
        struct Invariant: std::integral_constant<bool, Hashable<T>::value && boost::has_less<T, T, bool>::value && boost::has_equal_to<T, T, bool>::value> {};
```

Downside is compiler error messages are probably worse than the boost concept checking variant. Assuming I didn't miss a way to avoid the default construction assumption, would you prefer I keep the `Hashable` concept as-is with default-construction or refactor it with `type_traits`?

include/boost/graph/isomorphism.hpp
jeremy-murphy: Ah, no, you're right, it would be silly to do it the way I suggested.

jeremy-murphy:
Can you merge `develop` into your branch to get the updated CI configuration please?

include/boost/graph/isomorphism.hpp
jeremy-murphy: Ultimately, I think we shouldn't be defining a `Hashable` concept in an `isomorphism` algorithm file, we should be relying on an existing concept defined either in the standard library or a Boost concept library.
I think that can be done in the future as part of a broader modernization change to Boost.Graph, so for now, I'm thinking just take the Hashable concept stuff out, don't worry about it.

jeremy-murphy: And the CI tests are failing because `concept` is now a reserved word.  :)

include/boost/graph/isomorphism.hpp
```diff
@@ -158,46 +173,66 @@ namespace detail
...
+            assert(std::is_sorted(first, last));
+            multiplicity_map invar_multiplicity;
+
+            typedef typename std::vector< invariant_t >::const_iterator invar_iter;
+            typedef typename multiplicity_map::iterator invar_map_iter;
+
+            if(first == last) 
+                return invar_multiplicity;
+            
+            invar_iter it = std::move(first);
```
jeremy-murphy, line 189: Ahh, not sure what you're thinking here? Just keep using `first`, no need to copy/move it.

jargonzombies: I always thought range algorithms modifying `first` weren't naming their variables right :man_shrugging: I've removed it.

jeremy-murphy: I get what you mean, but I guess one way to think about it is that it's named for what the user passes in. What you do with it inside the algorithm is up to you. And consider, what is the meaning of `it`?

jargonzombies: That's fair enough, `it` isn't the best variable name either. Could be short for iterator or item, which doesn't help all that much. `iter` might have been better. In any case, I think it's fine the way it is now.

jeremy-murphy:
I've forgotten where I was up to on this.  :\

include/boost/graph/isomorphism.hpp
```diff
@@ -158,46 +161,63 @@ namespace detail
...
+            invar1_array.push_back(invariant1(v));
+            sort(invar1_array);
+
+            std::vector< invariant_t > invar2_array;
+            invar2_array.reserve(num_vertices(G2));
+            BGL_FORALL_VERTICES_T(v, G2, Graph2)
+            invar2_array.push_back(invariant2(v));
+            sort(invar2_array);
+            if (!equal(invar1_array, invar2_array))
+                return false;
```
jeremy-murphy, line 214: Are the number of vertices in G1 and G2 guaranteed to be equal here? 

If not, then one minor optimization we can do is to check that the number of vertices in both graphs is equal before going ahead with creating the arrays and sorting them, etc.

include/boost/graph/isomorphism.hpp
```diff
@@ -158,46 +161,63 @@ namespace detail
...
+                else
+                {
+                    invar = *first;
+                    inserted = invar_multiplicity.emplace(invar, 1).first;
+                }
+            }
+
+            return invar_multiplicity;
+        }
+
```
jeremy-murphy, line 194: As I mentioned in another comment, I think you can delete this function.

include/boost/graph/isomorphism.hpp
```diff
@@ -39,8 +39,8 @@ namespace detail
         typedef typename graph_traits< Graph2 >::vertex_descriptor vertex2_t;
         typedef typename graph_traits< Graph1 >::edge_descriptor edge1_t;
         typedef typename graph_traits< Graph1 >::vertices_size_type size_type;
-        typedef typename Invariant1::result_type invar1_value;
-        typedef typename Invariant2::result_type invar2_value;
+        typedef typename Invariant1::result_type invariant_t;
+        typedef unordered_map< invariant_t, size_type > multiplicity_map;
```
jeremy-murphy, line 43: It's important to let the user choose this type, so it needs to become a template parameter with a default value, like:
```
    template < typename Graph1, typename Graph2, typename IsoMapping,
        typename Invariant1, typename Invariant2, typename IndexMap1,
        typename IndexMap2, typename InvariantCountMap = boost::unordered_flat_map< invariant_t, size_type > >
```
Note that I've used `boost::unordered_flat_map`, which generally blows `boost::unordered_map` out of the water. (See [here](https://www.boost.org/doc/libs/develop/libs/unordered/doc/html/unordered.html).)

include/boost/graph/isomorphism.hpp
```diff
@@ -81,17 +81,20 @@ namespace detail
...
             }
             bool operator()(const vertex1_t& x, const vertex1_t& y) const
             {
-                return multiplicity[invariant1(x)]
-                    < multiplicity[invariant1(y)];
+                auto x_multiplicity_iter = multiplicity->find(invariant1(x));
+                assert(x_multiplicity_iter != multiplicity->end());
+                auto y_multiplicity_iter = multiplicity->find(invariant1(y));
+                assert(y_multiplicity_iter != multiplicity->end());
+                return *x_multiplicity_iter < *y_multiplicity_iter;
```
jeremy-murphy, line 94: I'd like to revert this back to the original code and add a TODO comment about a generic assertion that the map already contains the elements.

jeremy-murphy:
@jan-grimo , do you still have time to finish this off? I'd like to get it done before I merge `develop` into `master`. I think there's just a bit of minor editing to be done.

include/boost/graph/isomorphism.hpp
jan-grimo: Yes, this is guaranteed by [an earlier check](https://github.com/boostorg/graph/blob/b737cca58ae6d31cdd4241ff4de7b4755ba614a7/include/boost/graph/isomorphism.hpp#L596).

include/boost/graph/isomorphism.hpp
```diff
@@ -158,46 +161,63 @@ namespace detail
...
             BGL_FORALL_VERTICES_T(v, G1, Graph1)
             V_mult.push_back(v);
-            {
-                std::vector< size_type > multiplicity(max_invariant, 0);
-                BGL_FORALL_VERTICES_T(v, G1, Graph1)
-                ++multiplicity.at(invariant1(v));
-                sort(
-                    V_mult, compare_multiplicity(invariant1, &multiplicity[0]));
-            }
+            sort(V_mult, compare_multiplicity(invariant1, multiplicities(invar1_array.begin(), invar1_array.end())));
```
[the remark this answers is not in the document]
jan-grimo: Are you sure? We discussed the performance merits of using the `multiplicities` function [here](https://github.com/boostorg/graph/pull/349#discussion_r1324068860) before.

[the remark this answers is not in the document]
jeremy-murphy: Thanks for the reminder.  I have to laugh at my memory's frailty.  :)

include/boost/graph/isomorphism.hpp
jeremy-murphy: Actually, don't worry about this after all. Apologies if you have already started working on the change.

include/boost/graph/isomorphism.hpp
jeremy-murphy: This is the only change left, then it's good to merge!

jan-grimo:
I don't understand what's wrong :see_no_evil:

```
2024-04-12T16:13:34.8397092Z D:\a\graph\boost-root\boost/unordered/detail/foa/core.hpp(1284): error C2146: syntax 
error: missing '>' before identifier 'key_type'
2024-04-12T16:13:34.8410724Z D:\a\graph\boost-root\boost/unordered/detail/foa/core.hpp(1289): note: see reference to class template instantiation 'is_map<Container>' being compiled
2024-04-12T16:13:34.8528000Z 
```

jeremy-murphy:
> I don't understand what's wrong 🙈
> 
> ```
> 2024-04-12T16:13:34.8397092Z D:\a\graph\boost-root\boost/unordered/detail/foa/core.hpp(1284): error C2146: syntax 
> error: missing '>' before identifier 'key_type'
> 2024-04-12T16:13:34.8410724Z D:\a\graph\boost-root\boost/unordered/detail/foa/core.hpp(1289): note: see reference to class template instantiation 'is_map<Container>' being compiled
> 2024-04-12T16:13:34.8528000Z 
> ```

Strange, I don't know either.

It may simply be that MSVC 14.0 is broken and I should take it out of the list.

jeremy-murphy:
I removed it, so please merge develop into your branch and try again.

jeremy-murphy:
Thanks for persisting with it, @jan-grimo !

jan-grimo:
You're welcome! Thank you for helping me get it over the line @jeremy-murphy

jeremy-murphy:
Are you able to quantify how the space efficiency changed? I mean, previously it was allocating a sequence container (vector) of size `max_invariant`, and now it allocates an associative container (hash map) of the same size as G1 -- did that change the overall space efficiency? Can you describe under what conditions the improved space efficiency occurs? I mean, what would previously cause `max_invariant` to be much greater than the size of G1 and thus incur a large allocation?

I just want to accurately document the change in the release notes. Thanks!

jan-grimo:
Space efficiency hasn't changed for the general use case without custom invariants or the already feasible approaches with custom invariants that output a contiguous or small output domain on 0..N. But now arbitrary vertex invariants over vertex properties are feasible without incurring enormous memory allocations.

I had a use case with many vertex properties where I had set up a perfect hash function as an invariant, but then ran into the hidden memory allocation, which in the case of a 64-bit hash caused an OOM abort. I circumvented it by mapping the hashes for both graphs from their distribution over 0..2^64-1 down onto 0..N (where N is the number of unique hashes for the two graphs being compared). But I was irked by the hidden allocation and the odd invariant functor interface and thought it could be easier, hence the MRs that came from #203.

jeremy-murphy:
Cool, thanks.

PR #350, opened 2023-09-08
Add a missing include to topology.hpp

vslashg:
boost::shared_ptr is used in this file without the corresponding include, so currently this file does not compile on its own.


jeremy-murphy:
Thanks for catching and fixing that!

PR #352, opened 2023-09-25
Remove link against boost regex library, as normally it is header only when using C++ >= 11.

Romain-Geissler-1A:
It seems you recently dropped all support for gcc 4.x and C++ < 11 in general, so normally with the compilers you still support the Boost regex library shall be header only. Thus avoid linking against it explicitly.

I didn't test this locally, so let's wait the result of the CI.

### DIFF
--- a/build/Jamfile.v2
+++ b/build/Jamfile.v2
@@ -16,7 +16,6 @@ lib boost_graph
     read_graphviz_new.cpp
     graphml.cpp
     :
-    <library>../../regex/build//boost_regex
     <link>shared:<define>BOOST_GRAPH_DYN_LINK=1
     # Without these flags, MSVC 7.1 and 8.0 crash
     # User reports that VC++ 8.0 does not fail anymore, so that is removed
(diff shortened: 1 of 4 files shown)


Romain-Geissler-1A:
It looks like this breaks all clang < 4.x tests. How ok is it to remove clang 3.6 to 3.9 from the supported compilers ? Clang 4.0.0 debuted in march 2017, so 6 years ago. The oldest gcc release you still support, gcc 5, debuted in April 2015, so 8 years ago.

jeremy-murphy:
Thanks for updating this.

I'd like to remove more old compilers from the CI, but I'll just check my notes to see if anyone was actively using them.

jeremy-murphy:
I just removed Clang 3.x from the Drone config, so please rebase on to the latest `develop`.

jeremy-murphy:
Thanks, that's great. The change looks good to me, but I'm not an expert on the b2 build configuration, so I might ask someone else to quickly review it too.

Romain-Geissler-1A:
Ok, I have reverted the CMakeLists.txt changes.

Romain-Geissler-1A:
Ping

jeremy-murphy:
> Ping

Sorry, been on holidays!

jeremy-murphy:
Thanks for improving that!

PR #354, opened 2023-11-17
graph/labeled_graph: added remove_labeled_vertex

DeuceBox:
I took this patch from this ticket:
https://svn.boost.org/trac/boost/ticket/9493
which no longer exists.

But I believe the original ticket issue was identical to https://github.com/boostorg/graph/issues/167 which this patch fixes.

### DIFF
--- a/include/boost/graph/labeled_graph.hpp
+++ b/include/boost/graph/labeled_graph.hpp
@@ -260,6 +260,54 @@ namespace graph_detail
     }
     //@}
 
+    /** @name Remove Labeled Vertex */
+    //@{
+    // Tag dispatch on random access containers (i.e., vectors)
+    template <typename Container, typename Label, typename Graph>
+    void remove_labeled_vertex(Container& c, Graph& g, Label const& l,
+        random_access_container_tag)
+    {
+        if (l < c.size())
+        {
+            boost::remove_vertex(c[l], g);
+            c.erase(c.begin() + l);
+        }
+    }
+
+    // Tag dispatch on multi associative containers (i.e. multimaps).
+    template <typename Container, typename Label, typename Graph>
+    void remove_labeled_vertex(Container& c, Graph& g, Label const& l,
+        multiple_associative_container_tag)
+    {
+        typename Container::iterator c_it = c.find(l);
+        if (c_it != c.end())
+        {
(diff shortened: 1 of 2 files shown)


jeremy-murphy:
I'm looking on my phone so I might be wrong, but it looks like it's running GCC 4.X in the CI, which I thought I removed. Are you up to date on develop?

DeuceBox:
> I'm looking on my phone so I might be wrong, but it looks like it's running GCC 4.X in the CI, which I thought I removed. Are you up to date on develop?

You're correct that my branch was quite far behind develop. I just rebased on the latest develop, pushed again, and the tests were still failing. It appears that the original patch I preserved from the original svn ticket/tracker only accounted for unique associative containers, and not a random access / vector container that the tests use. 

I did my best to update the patch and mimic the tag dispatch approach used by other methods to add support to `remove_labeled_vertex` for random access containers and multiple associative containers as well, and pushed that updated branch now,

jeremy-murphy:
Looks good. Could you please modify an existing unit test for labelled graphs to assert that the change is effective?

I.e., change a unit test so that it _fails_ without this change, then put it together with your changes so that it passes.

DeuceBox:
I've added a test for removing a labeled vertex, based on the original https://github.com/boostorg/graph/issues/167 post's code snippet. I tested that it fails without my patch, and passes with my patch.

jeremy-murphy:
> I've added a test for removing a labeled vertex, based on the original #167 post's code snippet. I tested that it fails without my patch, and passes with my patch.

That's great. 

It does make me wonder about duplicate labels, though -- what happens? Is it even possible? If duplicate labels are possible in any circumstance, then we need a test to confirm what happens.

DeuceBox:
Added a test that uses a multi associative container and adds two vertices with the same label, and then removes them one by one, asserting the graph vertex count at each stage.

jeremy-murphy:
> Added a test that uses a multi associative container and adds two vertices with the same label, and then removes them one by one, asserting the graph vertex count at each stage.

Fantastic. As soon as the CI is green, I'll merge it.

jeremy-murphy:
Thanks so much for fixing a long-standing bug!

PR #361, opened 2024-01-26
Update Boost.Filesystem usage

Lastique:
This commit updates tests to avoid using Boost.Filesystem APIs that were deprecated and then removed.


jeremy-murphy:
Any idea what's going on with GCC 9 on Linux?

Lastique:
If you mean Drone then this is a `b2` issue [here](https://github.com/bfgroup/b2/issues/360). Reportedly fixed (but not yet merged to Boost).

jeremy-murphy:
OK, cool. Thanks for this!

PR #362, opened 2024-01-27
Fix documentation hyperlink

sehe:
simple copy and paste error - I hope the HTML is the right place to fix


jeremy-murphy:
Thanks, I actually just merged the same fix from someone else!

PR #370, opened 2024-04-12
Fix typo in name adjacenct_vertices.

### LINKED ISSUE #188: Typo in name adjacenct_vertices()

jcdong98:
Fix #188 and #189.

Also add graph concept checks for labeled graph to avoid this kind of typo.


jeremy-murphy:
Cool, thank you! I think I should be able to merge this as soon as the checks complete.

Issue #364, opened 2024-02-18
Stack-overflow in boost::read_graphviz_detail::parser::parse_subgraph


jeremy-murphy:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=66719

```
+----------------------------------------Release Build Stacktrace----------------------------------------+
Command: /mnt/scratch0/clusterfuzz/resources/platform/linux/unshare -c -n /mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_boost_fa513528a55ae19c064fe12b99e08e0dbc98b83f/revisions/boost_graph_graphviz_fuzzer -rss_limit_mb=2560 -timeout=60 -runs=100 /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-ec3a74d351f80590a35c0c6bace0c3a2fd548235
Time ran: 0.5343647003173828
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 49547342
INFO: Loaded 1 modules   (10110 inline 8-bit counters): 10110 [0x7acd98, 0x7af516),
INFO: Loaded 1 PC tables (10110 PCs): 10110 [0x70bd78,0x733558),
/mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_boost_fa513528a55ae19c064fe12b99e08e0dbc98b83f/revisions/boost_graph_graphviz_fuzzer: Running 1 inputs 100 time(s) each.
Running: /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-ec3a74d351f80590a35c0c6bace0c3a2fd548235
AddressSanitizer:DEADLYSIGNAL
=================================================================
==38477==ERROR: AddressSanitizer: stack-overflow on address 0x7fff2c367df8 (pc 0x0000004c8a11 bp 0x7fff2c368640 sp 0x7fff2c367e00 T0)
SCARINESS: 10 (stack-overflow)
    #0 0x4c8a11 in __interceptor_memcpy /src/llvm-project/compiler-rt/lib/sanitizer_common/sanitizer_common_interceptors.inc:827:5
    #1 0x45602f in Set /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDictionary.h:32:5
    #2 0x45602f in FixedWord /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDictionary.h:26:43
[...]

PR #375, opened 2024-05-03
Refactoring graphviz_test for review

sehe:
(note this PR builds on #374)

The code was mostly fine (except for unhygienic `using namespace` in
places), but it was hard to see what was covered.

I've seperated fixtures (sample input + expected output), and models into separate namespaces.

This simplifies invocations of the ComparisonDriver (test_graph).

### DIFF
--- a/test/graphviz_test.cpp
+++ b/test/graphviz_test.cpp
@@ -17,6 +17,11 @@
 #include <boost/assign/std/map.hpp>
 #include <boost/graph/adjacency_list.hpp>
 #include <boost/graph/compressed_sparse_row_graph.hpp>
+#include <boost/core/lightweight_test.hpp>
+#include <boost/property_map/function_property_map.hpp>
+
+// SEHE: Apparently unused or indirectly included
+#include <boost/regex.hpp>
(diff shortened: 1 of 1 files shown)


sehe:
Updated for review comments https://github.com/boostorg/graph/issues/364#issuecomment-2100506381

test/graphviz_test.cpp
```diff
@@ -1,151 +1,192 @@
...
+typedef std::pair< node_t, node_t > edge_t;
 
-    bool operator()(T l, T r) const {
-        return std::abs(l - r) <=
-            (std::max)(f_ * (std::max)(std::abs(l), std::abs(r)), T());
-    }
+typedef float Mass;
+typedef double Weight;
+typedef std::map< node_t, Mass > expected_masses_t;
+typedef std::map< edge_t, Weight > expected_weights_t;
```
jeremy-murphy, line 31: Personally, I think the type names should be more like `node_mass_map_t`, `edge_weight_map_t` and the variables should be named `expected_mass`, etc, but I'm not going to hold up the review over it.  :)

Issue #378, opened 2024-06-10
creation of nested subgraphs by vertex descriptors is broken/unintuitive


reuther-genisys:
When using `subgraph::create_subgraph(VertexIterator first, VertexIterator last)`, the vertex descriptors are interpreted as vertex descriptors into the root graph. (`add_vertex` expects the given descriptor to be a root graph descriptor).

My expectation is that the vertex descriptors should be interpreted as descriptors for the subgraph on which `create_subgraph` is called on. This way you can get a subgraph object, do some vertex selection logic on it and pass the set of selected vertices to `create_subgraph` without ever needing to worry where the graph is situated in the subgraph tree.

The fix is to replace line 197 `add_vertex(*first, *m_children.back());` with `add_vertex(local_to_global(*first), *m_children.back());`.

jeremy-murphy:
I'm a little nervous about changing behaviour in a way that might break existing code, although I agree that the current implementation does not sound ideal.

Do you mind giving a more full example that demonstrates how unintuitive the current implementation is?

reuther-genisys:
The current approach breaks everywhere a graph is recursively split into smaller subgraphs.

My application is the decomposition of a cycle-free graph into a complete set of longest-possible traversal paths. (Complete here means that every vertex belongs to exactly one resulting traversal path). The approach is recursive and looks like this:

```
using Graph = boost::subgraph<G>;
using VertexD = boost::graph_traits<Graph>::vertex_descriptor;
using Path = std::vector<VertexD>;

void recurse(graph graph, std::vector<Path>& paths)
{
  if (graph.empty())
    return;

  std::vector<std::set<VertexD> > components = ComponentDecomposition(); // essentially connected_components(), sorting the result into sets for ease of use

  for (auto&& component: components)
  {
    Graph componentGraph = graph.create_subgraph(component.begin(), component.end()); 

    auto path = GetLongestPath(component);  // essentially bfs starting at each vertex of degree 1
    paths.push_back(path);

    std::set<VertexD> remainingVertices(vertices(component).first, vertices(component).second);
    for (auto&& v: path) remainingVertices.erase(path);

    recurse(component.create_subgraph(remainingVertices.begin(), remainingVertices.end()), paths);
}
```

Both `create_subgraph` calls here operate on vertex descriptors obtained from the present graph. With the current behavior, this code breaks immediately.

If you do not want to change behavior, the documentation should at least be updated (for all versions!) to reflect the behavior: 

[...]

PR #379, opened 2024-06-10
Fix creation of nested subgraphs by vertex descriptors

reuther-genisys:
When creating subgraphs on deeper levels of the subgraph tree, do not nterpret the vertex descriptors as global, but instead as vertex descriptors to the current graph.

Fix to https://github.com/boostorg/graph/issues/378

### DIFF
--- a/include/boost/graph/subgraph.hpp
+++ b/include/boost/graph/subgraph.hpp
@@ -194,7 +194,7 @@ public:
         m_children.back()->m_parent = this;
         for (; first != last; ++first)
         {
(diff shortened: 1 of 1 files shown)


jeremy-murphy:
To start with, can you just merge `develop` in (or rebase)? Thanks. I thought I would be able to with one button click, but seems I can't.

reuther-genisys:
> To start with, can you just merge `develop` in (or rebase)? Thanks. I thought I would be able to with one button click, but seems I can't.

Done.

PR #390, opened 2024-10-08
fixed incorrect callbacks to finish_edge in undirected_dfs

### LINKED ISSUE #389: Bug in `undirected_dfs` causing incorrect callbacks to `finish_edge` of `dfs_visitor`
## Minimal working example
<details>
<summary>Code</summary>

```c++
#include <iostream>

#include <boost/graph/undirected_dfs.hpp>

using Graph = boost::adjacency_list<
    boost::vecS,
    boost::vecS,
    boost::undirectedS,
    boost::no_property,
    boost::property<boost::edge_color_t, boost::default_color_type>>;

using Vertex = boost::graph_traits<Graph>::vertex_descriptor;
using Edge = boost::graph_traits<Graph>::edge_descriptor;

struct DFSVisitorLogger : boost::default_dfs_visitor {
  void log_vertex(const Vertex v, const std::string &event) {
    std::cout << "vertex " << v << " " << event << "\n";
  }
  void log_edge(const Edge e, const std::string &event, const Graph &g) {
    std::cout << "edge (" << boost::source(e, g) << "," << boost::target(e, g) << ") " << event << "\n";
  }

  void discover_vertex(Vertex v, const Graph &g) {
    log_vertex(v, "discovered");
  }
  void finish_vertex(Vertex v, const Graph &g) {
    log_vertex(v, "finished");
  }
  void examine_edge(Edge e, const Graph &g) {
    log_edge(e, "examined", g);
  }
  void tree_edge(Edge e, const Graph &g) {
    log_edge(e, "tree", g);
  }
  void back_edge(Edge e, const Graph &g) {
    log_edge(e, "back", g);
  }
  void forward_or_cross_edge(Edge e, const Graph &g) {
    log_edge(e, "forward_cross", g);
  }
  void finish_edge(Edge e, const Graph &g) {
    log_edge(e, "finished", g);
  }
};

int main() {
  Graph g(3);
  boost::add_edge(0, 1, g);
  boost::add_edge(1, 2, g);

  DFSVisitorLogger dfs_visitor_logger;
  boost::undirected_dfs(g,
      boost::visitor(dfs_visitor_logger)
      .edge_color_map(boost::get(boost::edge_color, g)));
}
```
</details>

Example Graph
```
0 (start) -- 1 -- 2
```

Actual output:
```
vertex 0 discovered
edge (0,1) examined
edge (0,1) tree
vertex 1 discovered
edge (1,0) examined
edge (1,0) finished
edge (1,2) examined
edge (1,2) tree
vertex 2 discovered
edge (2,1) examined
edge (2,1) finished
vertex 2 finished
edge (1,2) finished
vertex 1 finished
edge (1,2) finished   <-- should be "edge (0, 1) finished"
vertex 0 finished
edge (0,1) finished   <-- should not appear
```

Expected output:
```
vertex 0 discovered
edge (0,1) examined
edge (0,1) tree
vertex 1 discovered
edge (1,0) examined
edge (1,0) finished
edge (1,2) examined
edge (1,2) tree
vertex 2 discovered
edge (2,1) examined
edge (2,1) finished
vertex 2 finished
edge (1,2) finished
vertex 1 finished
edge (0,1) finished
vertex 0 finished
```

## Problem
The problem is that in [`undirected_dfs.hpp:61`](https://github.com/boostorg/graph/blob/ba3022a97d6ee0c7520425b3fa5a45404efb7908/include/boost/graph/undirected_dfs.hpp#L61) the following invariant for the entries `VertexInfo& back = (u, (src_e, out_e))` in the `stack` is established:
- `u`: vertex
- `src_e`: edge used to discover vertex `u` (i.e. `u` is the **target** of `src_e`)
- `out_e`: set of edges adjacent to `u` not yet examined
```
                 --out_e-->
                /
--src_e--> (u)  ---out_e-->
                \
                 --out_e-->
```
However, [`undirected_dfs.hpp:82`](https://github.com/boostorg/graph/blob/ba3022a97d6ee0c7520425b3fa5a45404efb7908/include/boost/graph/undirected_dfs.hpp#L82) violates this invariant by pointing `src_e` to the next edge `*ei = (u, v)` in the previous line 81 (i.e. `u` is the **source** vertex) and then pushes `(u, (src_e, out_e))` onto the stack, which corresponds to: 
```
                 --src_e-->
                /
--  .  --> (u)  ---out_e-->
                \
                 --out_e-->
```

## Solution
See pull request #390

## Related issues
Issue related to #222 and pull request #129, but the solution proposed there does not properly establish the described invariant.

danielyxyang:
Resolves #389

### DIFF
--- a/include/boost/graph/undirected_dfs.hpp
+++ b/include/boost/graph/undirected_dfs.hpp
@@ -79,7 +79,7 @@ namespace detail
                 {
                     vis.tree_edge(*ei, g);
                     stack.push_back(std::make_pair(u,
-                        std::make_pair(src_e, std::make_pair(ei+1, ei_end))));
+                        std::make_pair(src_e, std::make_pair(std::next(ei), ei_end))));
                     u = v;
                     src_e = *ei;
                     put(vertex_color, u, Color::gray());
--- a/include/boost/graph/undirected_dfs.hpp
+++ b/include/boost/graph/undirected_dfs.hpp
@@ -78,10 +78,10 @@ namespace detail
                 if (v_color == Color::white())
                 {
                     vis.tree_edge(*ei, g);
-                    src_e = *ei;
                     stack.push_back(std::make_pair(u,
-                        std::make_pair(src_e, std::make_pair(++ei, ei_end))));
+                        std::make_pair(src_e, std::make_pair(ei+1, ei_end))));
                     u = v;
+                    src_e = *ei;
                     put(vertex_color, u, Color::gray());
                     vis.discover_vertex(u, g);
                     boost::tie(ei, ei_end) = out_edges(u, g);
--- a/test/undirected_dfs_visitor.cpp
+++ b/test/undirected_dfs_visitor.cpp
@@ -1,0 +1,104 @@
+//=======================================================================
+// Copyright 2024
+// Author: Daniel Yang
+//
+// Distributed under the Boost Software License, Version 1.0. (See
+// accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+//=======================================================================
+
+#include <vector>
+#include <string>
+
+#include <boost/config.hpp>
+#include <boost/core/lightweight_test.hpp>
+
+#include <boost/graph/adjacency_list.hpp>
+#include <boost/graph/undirected_dfs.hpp>
+
+using Graph = boost::adjacency_list<
+        boost::vecS,
+        boost::vecS,
+        boost::undirectedS,
+        boost::no_property,
+        boost::property<boost::edge_color_t, boost::default_color_type>>;
+
+using Vertex = boost::graph_traits<Graph>::vertex_descriptor;
+using Edge = boost::graph_traits<Graph>::edge_descriptor;
+
+struct DFSVisitorLogger : boost::default_dfs_visitor {
+    std::vector<std::string> &log;
+
+    DFSVisitorLogger(std::vector<std::string> &log) : log(log) {}
+
+    void log_vertex(const Vertex v, const std::string &event) {
+        log.push_back("vertex " + std::to_string(v) + " " + event);
+    }
+    void log_edge(const Edge e, const std::string &event, const Graph &g) {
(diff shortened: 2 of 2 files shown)


jeremy-murphy:
Not sure why all those Drone checks failed, but you probably should update to the tip of `develop` anyway because I have fixed some CI configuration.

jeremy-murphy:
Please add a unit test that would fail on `develop` as it is but passes with your proposed change.

include/boost/graph/undirected_dfs.hpp
```diff
@@ -78,10 +78,10 @@ namespace detail
                 if (v_color == Color::white())
                 {
                     vis.tree_edge(*ei, g);
-                    src_e = *ei;
                     stack.push_back(std::make_pair(u,
-                        std::make_pair(src_e, std::make_pair(++ei, ei_end))));
+                        std::make_pair(src_e, std::make_pair(ei+1, ei_end))));
```
jeremy-murphy, line 82: Probably has to be this so that it works with edge iterators that aren't random access.
```suggestion
                        std::make_pair(src_e, std::make_pair(std::next(ei), ei_end))));
```
And, call me fussy, but I prefer this style anyway.

danielyxyang: Ah yes, makes sense! I'm not very fluent in C++, so thanks for pointing this out :)

jeremy-murphy:
HUh, that's weird, the GitHub actions CI doesn't seem to be running, and the Drone CI ran instantly.  :\
Sorry, I'll have to investigate what's going on with the CI before I can merge any PRs.

jeremy-murphy:
Thanks again!

jeremy-murphy:
I was happy with the implementation, so I merged it, but actually the test could do with some improvement if you have time.
Using a string as the output type works, but it's a little fragile. What would be better is a scoped enum, but since the output could be a vertex or an edge, it's not simple how to represent it. I would lean towards a sum type, but I'm open to alternative suggestions. Feel free to discuss ideas before you implement, if you're willing to take it on. 
I was also trying to think of more tests to add, and so far the only slightly meaningful test case I can think of is a V-shaped graph with three vertices. If the search starts from the obvious root, then the test will confirm that one branch is finished before the other one is started. It's not really related to the original bug anymore, it's just an edge case that I thought is worth having a test for.

Issue #388, opened 2024-10-04
"is_straight_line_drawing()" is buggy for large graphs


Hermann-SW:
I experienced "is_straight_line_drawing()" returning <kbd>false</kbd> for valid straight line drawing of a maximal planar graph with 10million vertices first:  
https://lists.boost.org/Archives/boost/2024/10/257979.php

Later I was able to reduce to 1million vertices, and now to only three vertices with this simplified gist:  
https://gist.github.com/Hermann-SW/992b43eef29f1ef28c4bf2727f0a7c16

After the straight line drawing for the K₃ was computed, I did overwrite the determined coordinates for the three vertices with the coordinates of three vertices reported for initial 10million vertex graph. With that coordinates and the edges in same direction as in previous graph the problem is recreated fast:
```
hermann@7950x:~$ f=is_straight_line_drawing.recreate
hermann@7950x:~$ g++ -O3 -Wall -pedantic -Wextra $f.cpp -o $f
hermann@7950x:~$ ./is_straight_line_drawing.recreate
is_straight_line_drawing.recreate: is_straight_line_drawing.recreate.cpp:101: int main(int, char**): Assertion `is_straight_line_drawing(g, straight_line_drawing)' failed.
Aborted (core dumped)
hermann@7950x:~$ 
```

The problem is that "is_straight_line_drawing()" working on coordinates of std::size_t does call function using double, causing false report:
```
inline bool intersects(double x1, double y1, double x2, double y2, double a1,
     double b1, double a2, double b2, double epsilon = 0.000001)
{ 
```

There are plenty of intersection test algorithms that avoid doing division and report correct result for integer coordinates, eg:  
1. https://en.wikipedia.org/wiki/Line%E2%80%93line_intersection#Given_two_points_on_each_line_segment
2. https://www.dcs.gla.ac.uk/~pat/52233/slides/Geometry1x1.pdf#page=4

"bool intersects()" needs to be implemented with integer coordinates for correct response.

This is the graph from recreate gist:
```
    graph g(3);
    add_edge(0, 1, g);
    add_edge(2, 0, g);
    add_edge(1, 2, g);
```

And these are the coordinates:
```
    straight_line_drawing[0].x = 4143438;
    straight_line_drawing[0].y = 86426;
    straight_line_drawing[1].x = 4064945;
    straight_line_drawing[1].y = 7932;
    straight_line_drawing[2].x = 4064944;
    straight_line_drawing[2].y = 7931;
```

[...]

PR #391, opened 2024-10-12
Failing test for is_straight_line_drawing

Hermann-SW:
For #388

### DIFF
--- a/include/boost/graph/is_straight_line_drawing.hpp
+++ b/include/boost/graph/is_straight_line_drawing.hpp
@@ -16,75 +16,16 @@
 #include <boost/graph/properties.hpp>
 #include <boost/graph/planar_detail/bucket_sort.hpp>
 
+#include <boost/geometry/algorithms/crosses.hpp>
+#include <boost/geometry/geometries/linestring.hpp>
+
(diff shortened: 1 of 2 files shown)


Hermann-SW:
@jeremy-murphy 
I created this failing testcase because you asked for that in https://github.com/boostorg/graph/issues/388.
Will the pull request be accepted before fix is available?
Or only after a fix will be provided making CIs green again?

jeremy-murphy:
Only after a fix is provided that makes it green. I understand that you're excited about this topic, but you will have to be patient. I do this maintenance voluntarily, so it has to come in between my full-time work and parenting, which doesn't leave much.

jeremy-murphy:
I wonder why `CI / macos (clang)` is green when it should be red like the others...

jeremy-murphy:
Hermann, I hope you don't mind, but I pushed directly to your PR instead of creating another PR, etc.

jeremy-murphy:
Sorry, I won't push on to someone else's PR again like that. I thought it was the simplest thing to do at the time, but code is never as simple as I hope. :)   So I created a new PR with your commits in #412

PR #392, opened 2024-10-31
Split CMake identities: superproject and 'developer'

jeremy-murphy:
I prefer to use CMake as a build system, but the existing CMake files in Boost don't really help.

1. I don't want to treat every library as an active project when I'm only working on one.
2. I need the header files to be part of the project so that an IDE treats them as the source code of the project.

I'm always maintaining some split-identity CMake like this in order to do work, so I thought I may as well merge it upstream if 
1. It's compatible with super-project plans for CMake, and
2. other developers would find it useful.


jeremy-murphy:
This is still somewhat broken, I haven't got the tests to compile properly yet, I think it's partly missing the required Boost.UTF macros, but also some unexpected `static_assert` failures.

jeremy-murphy:
I actually don't know who else works on Boost.Graph _and_ prefers CMake, so I'll just tag a few recent contributors to see if they are interested: @jan-grimo @danielyxyang @brunom @sehe @jcdong98 @andrea-cassioli-maersk @vslashg

jeremy-murphy:
Peter, I assigned you because I assume you know about the super-project CMake business.

andrea-cassioli-maersk:
I do work with CMake usually/

pdimov:
This looks OK at a first glance, but you should change
```
if (BOOST_SUPERPROJECT_VERSION)
```
to
```
if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
```
which checks whether this is the root project (as opposed to checking whether it's part of the Boost superproject.)

In principle, Boost libraries are also usable without a superproject, as direct subprojects, although in this specific case this probably isn't very practical because of the number of dependencies.

pdimov:
Also, if you're going to make changes to CMakeLists.txt, you should first add CI jobs that verify that current uses aren't broken. Or I can add them for you if you prefer.

jeremy-murphy:
> Also, if you're going to make changes to CMakeLists.txt, you should first add CI jobs that verify that current uses aren't broken. Or I can add them for you if you prefer.

If you have some pre-canned test, then yes, please go ahead!

jeremy-murphy:
> This looks OK at a first glance, but you should change
> 
> ```
> if (BOOST_SUPERPROJECT_VERSION)
> ```
> 
> to
> 
> ```
> if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
> ```
> 
> which checks whether this is the root project (as opposed to checking whether it's part of the Boost superproject.)
> 
> In principle, Boost libraries are also usable without a superproject, as direct subprojects, although in this specific case this probably isn't very practical because of the number of dependencies.

OK, sure, will do. I see there are some new variables with `PROJECT` in their name specifically for this purpose, but they require a fairly recent version of CMake, so I'll just go with whatever is supported by our minimum requirement.

pdimov:
CI jobs for testing CMake subdir/install use added. I arbitrarily picked the `adj_list_edge_list_set.cpp` test as the `main.cpp` file in both cases.

jeremy-murphy:
> CI jobs for testing CMake subdir/install use added. I arbitrarily picked the `adj_list_edge_list_set.cpp` test as the `main.cpp` file in both cases.

Actually, I'm a bit confused -- where did you add them to?

jeremy-murphy:
@andrea-cassioli-maersk have you tried using this?

andrea-cassioli-maersk:
not really, but I can give it a shot time permitting

jeremy-murphy:
If you could, that would be awesome, because I would like to merge it but so far I'm the only person that has used it.

andrea-cassioli-maersk:
The Cmake itself seems to work, I get some compile errors like
```
      |                                                   ^
In file included from /Users/andrea.cassioli/workspace/graph/test/adj_matrix_cc.cpp:9:
/opt/homebrew/include/boost/graph/graph_concepts.hpp:142:13: error: use of undeclared identifier 'degree'
  142 |         n = degree(v, cg);
      |             ^

```

but it might be because I using the boost distribution form `brew` on OSX, could it be?

jeremy-murphy:
Thanks for having a look. Hmmmm, I can't think of a good reason why distribution via `brew` would introduce problems but anything is possible.

Becheler:
@jeremy-murphy Can I help somehow testing/merging it ? 😄

jeremy-murphy:
> @jeremy-murphy Can I help somehow testing/merging it ? 😄 

Yeah, just try it out! Tell me if there are any problems. 
I'm reasonably happy with it and always use it when I'm working on BGL so I've been meaning to simply make sure it's up to date and then merge it.

pdimov:
Breaks testing Boost, see https://github.com/boostorg/boost/actions/runs/28580529793/job/84739424140

jeremy-murphy:
> Breaks testing Boost, see https://github.com/boostorg/boost/actions/runs/28580529793/job/84739424140

Oh, that's a pity. Thanks for letting me know. If I can't fix it quickly, I'll just revert it.

PR #394, opened 2024-11-08
Fix missing namespace name

tsnorri:
For some reason when using `cycle_canceling` as part of a larger algorithm, Clang++19 reported that it could not find `cycle_canceling_dispatch1`. Compiling [the example](https://www.boost.org/doc/libs/1_86_0/libs/graph/example/cycle_canceling_example.cpp) with the same options did work. In any case I think the namespace name should be applied to the call.

### DIFF
--- a/include/boost/graph/cycle_canceling.hpp
+++ b/include/boost/graph/cycle_canceling.hpp
@@ -172,7 +172,7 @@ namespace detail
 template < class Graph, class P, class T, class R >
 void cycle_canceling(Graph& g, const bgl_named_params< P, T, R >& params)
 {
-    cycle_canceling_dispatch1(g,
+    detail::cycle_canceling_dispatch1(g,
         choose_const_pmap(get_param(params, edge_weight), g, edge_weight),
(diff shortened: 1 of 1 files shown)


jeremy-murphy:
I agree, thanks for the fix!

jeremy-murphy:
Actually... although on the surface it looks right, I was expecting the example to have something hacky in it to make it work, but it doesn't. So I would like to know why it doesn't work for you. Or, why it works for the example. It's just slightly too mysterious to leave unanswered.

tsnorri:
I’ll see if I can find out the reason.

tsnorri:
So I got the syntax tree from Clang and in the case of the example, `cycle_canceling_dispatch1` is found via ADL. The exact declaration of the function template specialisation is quite long but the passed `adjacency_list` has (indirectly) a template parameter whose type is `boost::detail::edge_desc_impl`. I think this causes `boost::detail` to be considered when the compiler looks for `cycle_canceling_dispatch1`.

To me this doesn’t seem to be intentional.

Issue #399, opened 2024-12-05
multiple bugs in maximum_weighted_matching()


jorisvr:
`maximum_weighted_matching()` gives incorrect answers for some graphs.
And the function reports failed assertions for some other graphs.
And the function segfaults for some other graphs as previously reported #199 .
And the function fails to terminate for some other graphs as previously reported #223 .

These issues are not rare. The failure rate appears to scale with size and density of the graph. I find that random graphs of 1000 vertices and 10000 edges with integer weights, cause the algorithm to fail for more than 50% of the graphs.

The algorithm is also fairly slow for graphs of 1000 or more vertices.

To clearly establish that these problems exist, I prepared a new testsuite for `maximum_weighted_matching()`. It contains a bunch of general test cases that are passing, as well as a bunch of test cases that specifically trigger each of the 4 types of failures in the algorithm. I will attach a pull request to this issue with the new tests.

A possible way to solve these issues would be to replace the function with a new implementation based on code that I have been working on here: https://git.jorisvr.nl/joris/maximum-weight-matching

That code would be faster, running in O(n * m * log n) compared to O(n^3) for the current algorithm. It is also more complicated. It needs a specific type of priority queue to achieve the complexity bound. Obviously this code does not fit in BGL in its current form. It would have to be rewritten to fit the BGL API and Boost coding conventions. I'd like to try that. But before I spend a lot of time on it, I would like to know whether there is any interest here in a replacing the algorithm. I understand of course that the new code would have to be reviewed and judged on its merits etc. But I'm not going to waste time on it if it is clear up front that there is no interest in this type of rework.

jeremy-murphy:
I'm very interested.

jeremy-murphy:
As you've discovered from looking through the issues, there are multiple problems with this algorithm, so it would be a top priority for me to see it fixed. 
Performance improvement would be icing on the cake. 
A complicated algorithm is no drama so long as there are plenty of tests.

jorisvr:
Ok great. I need some time to work on the code, then I will link a pull request here.

jorisvr:
Ok. I pushed a new implementation to the existing PR #400 .
I think it is ready. Any feedback is very welcome.

jeremy-murphy:
I've asked the people that opened the original bugs to confirm it is fixed for them, but you can close this one whenever you're ready.

jorisvr:
Confirmed that I no longer see issues after merging #400.

Issue #223, opened 2020-06-10
maximum_weighted_matching never terminates for some edge orders


pascalhpi:
For maximum_weighted_matching the order in which edges are added to the graph makes a difference. For one order it works perfectly, but for the other the algorithm never terminates. Here is a modified version of the example to show the problem:
```c++
//=======================================================================
// Copyright (c) 2018 Yi Ji
//
// Distributed under the Boost Software License, Version 1.0.
// (See accompanying file LICENSE_1_0.txt or copy at
// http://www.boost.org/LICENSE_1_0.txt)
//
//=======================================================================

#include <iostream>
#include <vector>
#include <string>
#include <boost/graph/adjacency_list.hpp>
#include <boost/graph/maximum_weighted_matching.hpp>

using namespace boost;

typedef property< edge_weight_t, int, property< edge_index_t, int > >
    EdgeProperty;
typedef adjacency_list< vecS, vecS, undirectedS, no_property, EdgeProperty >
    my_graph;

int main(int argc, const char* argv[])
{
    graph_traits< my_graph >::vertex_iterator vi, vi_end;
    const int n_vertices = 10;
    my_graph g(n_vertices);

    // vertices can be refered by integers because my_graph use vector to store
    // them

    // NOT WORKING
    add_edge(8, 2, EdgeProperty(10000-1002), g);
    add_edge(1, 4, EdgeProperty(10000-502), g);
    add_edge(9, 3, EdgeProperty(10000-1007), g);
    add_edge(2, 6, EdgeProperty(10000-1), g);
    add_edge(9, 7, EdgeProperty(10000-1002), g);
    add_edge(0, 5, EdgeProperty(10000-1504), g);
    add_edge(9, 4, EdgeProperty(10000-1006), g);
    add_edge(9, 5, EdgeProperty(10000-1005), g);
    add_edge(9, 6, EdgeProperty(10000-1004), g);
    add_edge(0, 6, EdgeProperty(10000-1505), g);
    add_edge(1, 7, EdgeProperty(10000-506), g);
    add_edge(0, 3, EdgeProperty(10000-1502), g);
    add_edge(0, 4, EdgeProperty(10000-1503), g);
    add_edge(2, 4, EdgeProperty(10000-3), g);
    add_edge(1, 6, EdgeProperty(10000-504), g);
    add_edge(8, 5, EdgeProperty(10000-1004), g);
    add_edge(1, 3, EdgeProperty(10000-501), g);
[...]

PR #400, opened 2024-12-05
Replace implementation of maximum_weighted_matching()

### LINKED ISSUE #199: maximum_weighted_matching() segfaults with constant weights
Some graphs cause maximum_weighted_matching() to segfault due to a stack overflow when unit weights are used.

Please see minimal example linked below which is a simple modification of the standard example from the library.
 
https://gist.github.com/count0/551867e10973d4b2dc047b8528908281

jorisvr:
This is a re-implementation of maximum_weighted_matching, based on a paper by Zvi Galil.
The new code runs in time O(V^3).
A new set of testcases are also added.

Resolves #199 #223 #399

The code has been tested extensively on large random graphs, using LEMON as a reference.

Faster algorithms are known for this problem. I initially planned to implement the O(V*E*log(V)) algorithm by Galil, Micali and Gabow. However it needs a mutable heap with features that are not readily available in the BGL, and it needs a special kind of mergeable priority queue. While possible, I feel that the amount of code would be disproportionate. So I decided to fall back to a simpler O(V^3) algorithm, essentially the same algorithm that inspired the previous implementation. 

Feedback is very welcome. I will already mention a few points that may draw criticism:
- I kept `brute_force_maximum_weighted_matching()` unchanged. This function is not very useful in my opinion, but it was part of the public API and there is no need to change it.
- The documented API of `maximum_weighted_matching()` is backwards compatible with the previous code. But I removed the class `weighted_augmenting_path_finder`, which was essentially an internal detail although it lived in the global `boost` namespace.
- I do runtime checks on some aspects of the input graph (vertex indices and range of edge weights). I don't see this much in the BGL and I guess it may be frowned upon. The thing is, the code will trigger undefined behavior if these preconditions are violated, and I feel like I can't let that happen.
- Once a matching has been computed, a separate (much faster) algorithm can verify that the matching is optimal. If the primary algorithm is correct, this verification will never fail. I enabled the verification step by default, even though it is redundant and never fails. The primary algorithm is tricky. I feel that the certainty provided by the verification step is worth more than the clock cycles it costs.

### DIFF
--- a/doc/maximum_weighted_matching.html
+++ b/doc/maximum_weighted_matching.html
@@ -1,5 +1,6 @@
 <html><head><!--
      Copyright 2018 Yi Ji
+     Copyright 2025 Joris van Rantwijk
 
      Use, modification and distribution is subject to the Boost Software
      License, Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
@@ -19,6 +20,9 @@ void maximum_weighted_matching(const Graph&amp; g, MateMap mate);
 
 template &lt;typename Graph, typename MateMap, typename VertexIndexMap&gt;
 void maximum_weighted_matching(const Graph&amp; g, MateMap mate, VertexIndexMap vm);
+
+template &lt;typename Graph, typename MateMap, typename VertexIndexMap, typename EdgeWeightMap, bool Verify = true&gt;
+void maximum_weighted_matching(const Graph&amp; g, MateMap mate, VertexIndexMap vm, EdgeWeightMap weights);
 
 template &lt;typename Graph, typename MateMap&gt;
 void brute_force_maximum_weighted_matching(const Graph&amp; g, MateMap mate);
@@ -56,77 +60,109 @@ maximum weighted matching in any undirected graph. The matching is returned in a
 <a href="../../property_map/doc/ReadWritePropertyMap.html">ReadWritePropertyMap</a>
 that maps vertices to vertices. In the mapping returned, each vertex is either mapped
 to the vertex it's matched to, or to <tt>graph_traits&lt;Graph&gt;::null_vertex()</tt> if it
-doesn't participate in the matching. If no <tt>VertexIndexMap</tt> is provided, both functions
-assume that the <tt>VertexIndexMap</tt> is provided as an internal graph property accessible
-by calling <tt>get(vertex_index, g)</tt>.
+doesn't participate in the matching.
+</p>
+
+<h3>Algorithm Description</h3>
 
 <p>
-The maximum weighted matching problem was solved by Edmonds in [<a href="bibliography.html#edmonds65:_max_weighted_match">74</a>].
-The implementation of <tt>maximum_weighted_matching</tt> followed Chapter 6, Section 10 of [<a href="bibliography.html#lawler76:_comb_opt">20</a>] and
-was written in a consistent style with <tt>edmonds_maximum_cardinality_matching</tt> because of their algorithmic similarity.
-In addition, a brute-force verifier <tt>brute_force_maximum_weighted_matching</tt> simply searches all possible matchings in any graph and selects one with the maximum weight sum.
-
-</p><h3>Algorithm Description</h3>
-
-Primal-dual method in linear programming is introduced to solve weighted matching problems. Edmonds proved that for any graph,
-the maximum number of edges in a matching is equal to the minimum capacity of an odd-set cover; this further enable us to prove a max-min duality theorem for weighted matching.
-Let <i>H<sub>k-1</sub></i> denote any graph obtained from G by contracting odd sets of three or more nodes and deleting single nodes,
-where the capacity of the family of odd sets (not necessarily a cover of G) is <i>k-1</i>. Let <i>X<sub>k</sub></i> denote any matching containing <i>k</i> edges.
-Each edge <i>(i, j)</i> has a weight <i>w<sub>ij</sub></i>. We have:
-<math>
-max<i><sub>X<sub>k</sub></sub></i> min {<i>w<sub>ij</sub>|(i, j) ϵ X<sub>k</sub></i>} = min<i><sub>H<sub>k-1</sub></sub></i> max {<i>w<sub>ij</sub>|(i, j) ϵ H<sub>k-1</sub></i>}.
-</math>
-This matching duality theorem gives an indication of how the matching problem should be formulated as a linear programming problem. That is,
-the theorem suggests a set of linear inequalities which are satisfied by any matching, and it is anticipated that these inequalities describe a convex polyhedron
-with integer vertices corresponding to feasible matchings.
+The maximum weighted matching problem was solved by Edmonds in
+[<a href="bibliography.html#edmonds65:_max_weighted_match">74</a>].
+Subsequently, Gabow [<a href="bibliography.html#gabow76">75</a>] and
+Lawler [<a href="bibliography.html#lawler76:_comb_opt">20</a>] developed methods
+to reduce the run time from <i>O(V<sup>4</sup>)</i> to <i>O(V<sup>3</sup>)</i>.
+The implementation of <tt>maximum_weighted_matching</tt> follows a description
+of the <i>O(V<sup>3</sup>)</i> algorithm by Galil [<a href="bibliography.html#galil86">77</a>].
+</p>
 
 <p>
-For <tt>maximum_weighted_matching</tt>, the management of blossoms is much more involved than in the case of <tt>max_cardinality_matching</tt>.
-It is not sufficient to record only the outermost blossoms. When an outermost blossom is expanded,
-it is necessary to know which blossom are nested immediately with it, so that these blossoms can be restored to the status of the outermost blossoms.
-When augmentation occurs, blossoms with strictly positive dual variables must be maintained for use in the next application of the labeling procedure.
+Starting from an empty matching, the algorithm repeatedly augments the matching until no further improvement is possible.
+Alternating trees and blossoms are used to find an augmenting path, in a way that is similar to maximum cardinality matching.
+A linear programming technique is used to ensure that the selected augmenting path has maximum weight.
+This involves assigning dual variables to vertices and blossoms, and computing slack values for edges.
+</p>
 
 <p>
-The outline of the algorithm is as follow:
+The outline of the algorithm is as follows:
+</p>
 
 <ol start="0">
-<li>Start with an empty matching and initialize dual variables as a half of maximum edge weight.</li>
-<li>(Labeling) Root an alternate tree at each exposed node, and proceed to construct alternate trees by labeling, using only edges with zero slack value.
+<li>Start with an empty matching and initialize dual variables to half of the maximum edge weight.</li>
+<li>(Labeling) Root an alternate tree at each non-matched vertex, and proceed to construct alternate trees by labeling, using only edges with zero slack value.
 If an augmenting path is found, go to step 2. If a blossom is formed, go to step 3. Otherwise, go to step 4. </li>
-<li>(Augmentation) Find the augmenting path, tracing the path through shrunken blossoms. Augment the matching,
-correct labels on nodes in the augmenting path, expand blossoms with zero dual variables and remove labels from all base nodes. Go to step 1.</li>
-<li>(Blossoming) Determine the membership and base node of the new blossom and supply missing labels for all non-base nodes in the blossom.
+<li>(Augmentation) Find the augmenting path, tracing the path through shrunken blossoms. Augment the matching.
+Deconstruct and unlabel the alternating trees traversed by the augmenting path.
+Go to step 1.</li>
+<li>(Blossoming) Identify the blossoms in the alternating cycle and shrink these into a newly formed blossom.
 Return to step 1.</li>
-<li>(Revision of Dual Solution) Adjust the dual variables based on the primal-dual method. Go to step 1 or halt, accordingly.</li>
+<li>(Updating dual variables) Adjust the dual variables based on the primal-dual method.
+If this enables further growth of the alternating trees, return to step 1.
+Otherwise, halt the algorithm; the matching has maximum weight.</li>
 </ol>
 
-Note that in <tt>maximum_weighted_matching</tt>, all edge weights are multiplied by 4, so that all dual variables always remain as integers if all edge weights are integers.
-Unlike <tt>max_cardinality_matching</tt>, the initial matching and augmenting path finder are not parameterized,
-because the algorithm maintains blossoms, dual variables and node labels across all augmentations.
+<p>
+The implementation deviates from the description in [<a href="bibliography.html#galil86">77</a>] on a few points:
+</p>
 
-The algorithm's time complexity is reduced from <i>O(V<sup>4</sup>)</i> (naive implementation of [<a href="bibliography.html#edmonds65:_max_weighted_match">74</a>])
-to <i>O(V<sup>3</sup>)</i>, by a delicate labeling procedure [<a href="bibliography.html#gabow76">75</a>] to avoid re-scanning labels after revision of the dual solution.
-Special variables <i>pi, tau, gamma</i> and two arrays <i>critical_edge, tau_idx</i> are introduced for this purpose.
-Please refer to [<a href="bibliography.html#lawler76:_comb_opt">20</a>] and code comments for more implementation details.
+<ul>
+<li>After augmenting the matching, blossoms with zero dual are not expanded right away.
+Expanding such blossoms is correct, but potentially unnecessary.
+If expansion is required, it will eventually happen by assigning label T and going through a zero-delta dual update.</li>
+<li>After augmenting, labels and alternating trees are reused for the next round.
+Labels are erased only from the two alternating trees that are touched by the augmenting path.
+Although this strategy does not change the worst-case time complexity of the algorithm, it appears to provide a significant speedup for random graphs.</li>
+<li>The algorithm keeps track of the minimum-slack edge <i>e<sub>k,l</sub></i> for each pair of blossoms.
+Storing these edges in a 2-dimensional array would increase the space complexity to <i>O(V<sup>2</sup>)</i>.
+This is avoided by storing each row of the array as a list <tt>best_edge_set</tt> that contains only the non-empty elements.</li>
+</ul>
+
+<p>
+If edge weights are integers, the algorithm uses only integer computations.
+This is achieved by internally multiplying edge weights by 2, which ensures that all dual variables are also integers.
+</p>
+
+<p>
+After computing the maximum matching, a verification step may be performed to check that the matching is optimal.
+The verification step is simpler and faster (<i>O(V<sup>2</sup>)</i>) than the search for the matching.
+It always succeeds unless there is a bug in the matching algorithm.
+Since verification checks for exact optimality, it can only be used if edge weights are integers.
+</p>
+
+<p>
+In addition, a brute-force implementation <tt>brute_force_maximum_weighted_matching</tt> is provided.
+This algorithm simply searches all possible matchings and selects one with the maximum weight sum.
+</p>
 
 </p><h3>Where Defined</h3>
 
 <p>
 <a href="../../../boost/graph/maximum_weighted_matching.hpp"><tt>boost/graph/maximum_weighted_matching.hpp</tt></a>
+</p>
 
-</p><h3>Parameters</h3>
+<h3>Parameters</h3>
 
 IN: <tt>const Graph&amp; g</tt>
 <blockquote>
 An undirected graph. The graph type must be a model of
 <a href="VertexAndEdgeListGraph.html">Vertex and Edge List Graph</a> and
 <a href="IncidenceGraph.html">Incidence Graph</a>.
-The edge property of the graph <tt>property_map&lt;Graph, edge_weight_t&gt;</tt> must exist and have numeric value type.<br>
+The graph may not contain parallel edges.
 </blockquote>
 
 IN: <tt>VertexIndexMap vm</tt>
 <blockquote>
-Must be a model of <a href="../../property_map/doc/ReadablePropertyMap.html">ReadablePropertyMap</a>, mapping vertices to integer indices.
+Must be a model of <a href="../../property_map/doc/ReadablePropertyMap.html">ReadablePropertyMap</a>,
+mapping vertices to integer indices in the range <tt>[0, num_vertices(g))</tt>.
+If this parameter is not explicitly specified, it is obtained from the internal vertex property of the graph
+by calling <tt>get(vertex_index, g)</tt>.
+</blockquote>
+
+IN: <tt>EdgeWeightMap weights</tt>
+<blockquote>
+Must be a model of <a href="../../property_map/doc/ReadablePropertyMap.html">ReadablePropertyMap</a>, mapping edges to weights.
+Edge weights must be integers or floating point values.
+If this parameter is not explicitly specified, it is obtained from the internal edge property of the graph
+by calling <tt>get(edge_weight, g)</tt>.
 </blockquote>
 
 OUT: <tt>MateMap mate</tt>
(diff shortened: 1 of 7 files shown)


jeremy-murphy:
Are the heap data structures in Boost.Heap not sufficient? They are mergeable and mutable.
However, even if they are sufficient, I would be happy to simply get a correct implementation first that fixes the bugs and get a fast implementation later. 
What do you think?

jeremy-murphy:
Thanks for the explanation. There's no problem with adding Boost.Heap as a dependency, as Boost.Graph already depends on many other parts of Boost. Sounds like the efficient algorithm would require adding a new data structure to Boost.Heap to start with, which shouldn't be too difficult, although I'm aware that the maintainer is not all that active any more. Given that the new implementation is much faster anyway, let's defer the efficient algorithm to later. Ultimately it would be nice to have a top-level algorithm that uses a heuristic to pick the fastest algorithm but users are still free to call specific algorithms. (Best of both worlds.)
Ok, now I have to find time to look over this code. Honestly, it could take a couple of weeks, so please be patient.
Thanks for the work!

jorisvr:
> Sounds like the efficient algorithm would require adding a new data structure to Boost.Heap to start with

I think the concatenable queue may be so special-purpose that it could just stay in BGL, but generalizing it is definitely also a valid option.

> Given that the new implementation is much faster anyway, let's defer the efficient algorithm to later.

Agreed. It occurs to me that the O(V E log(V)) algorithm also needs an `edge_index` property, which breaks backward compatibility. There may be ways to deal with this, but it seems like it will be a more difficult road than the current PR.

> Ok, now I have to find time to look over this code.

I understand. There is no hurry from my side. Thanks for supporting this effort.

jeremy-murphy:
What's the "nearest ancestor" problem referred to in the documentation for the fast Gabow algorithm? Is that LCA or something else?

jorisvr:
> What's the "nearest ancestor" problem referred to in the documentation for the fast Gabow algorithm?

To be honest, I don't know. I kept this comment from the documentation by Yi Ji as I saw no reason to remove it.

I know about the existence of that fast algorithm. I tried to read the paper by Gabow but I can not make heads or tails of it. Mehlhorn and Schaefer made an offhand remark that this algorithm may be unpractical (https://dl.acm.org/doi/10.1145/944618.944622 page 7) but that was a long time ago. I'm not aware of any public available implementation.

jeremy-murphy:
Doesn't surprise me too much. Bender et. al. made a similar remark about the theoretically optimal algorithm for LCA: it just ain't worth it.

jeremy-murphy:
Citation LCA remark: https://www.sciencedirect.com/science/article/abs/pii/S0196677405000854

doc/maximum_weighted_matching.html
```diff
@@ -20,6 +21,9 @@ <h1>
 template &lt;typename Graph, typename MateMap, typename VertexIndexMap&gt;
 void maximum_weighted_matching(const Graph&amp; g, MateMap mate, VertexIndexMap vm);
 
+template &lt;typename Graph, typename MateMap, typename VertexIndexMap, typename EdgeWeightMap, bool Verify = true&gt;
```
jeremy-murphy, line 24: Leave the Verify parameter out, it's not worth complicating the interface for. If a user really wants to verify every result, they can write a short wrapper for themselves.

doc/maximum_weighted_matching.html
```diff
@@ -56,77 +60,109 @@ <h1>
...
+<li>(Updating dual variables) Adjust the dual variables based on the primal-dual method.
+If this enables further growth of the alternating trees, return to step 1.
+Otherwise, halt the algorithm; the matching has maximum weight.</li>
 </ol>
 
-Note that in <tt>maximum_weighted_matching</tt>, all edge weights are multiplied by 4, so that all dual variables always remain as integers if all edge weights are integers.
-Unlike <tt>max_cardinality_matching</tt>, the initial matching and augmenting path finder are not parameterized,
-because the algorithm maintains blossoms, dual variables and node labels across all augmentations.
+<p>
+The implementation deviates from the description in [<a href="bibliography.html#galil86">77</a>] on a few points:
```
jeremy-murphy, line 104: I think you need to provide more explanation and evidence about why these deviations from an algorithm published in a refereed journal are good. Even better would be to get your deviations published in a journal. :)

doc/maximum_weighted_matching.html
```diff
@@ -140,24 +176,29 @@ <h3>Complexity</h3>
 
 <p>
 Let <i>m</i> and <i>n</i> be the number of edges and vertices in the input graph, respectively. Assuming the
-<tt>VertexIndexMap</tt> supplied allows constant-time lookup, the time complexity for
-<tt>maximum_weighted_matching</tt> is <i>O(n<sup>3</sup>)</i>. For <tt>brute_force_maximum_weighted_matching</tt>, the time complexity is exponential of <i>m</i>.
+<tt>VertexIndexMap</tt> and <tt>EdgeWeightMap</tt> both provide constant-time lookup, the time complexity for
+<tt>maximum_weighted_matching</tt> is <i>O(n<sup>3</sup>)</i>.
+For <tt>brute_force_maximum_weighted_matching</tt>, the time complexity is exponential of <i>m</i>.
```
jeremy-murphy, line 181: I could be mistaken, but I think the correct phrasing is "exponential _in_" rather than "exponential _of_".

test/Jamfile.v2
```diff
@@ -114,6 +114,7 @@ alias graph_test_regular :
     [ run king_ordering.cpp ]
     [ run matching_test.cpp ]
     [ run weighted_matching_test.cpp ]
+    [ run weighted_matching_test2.cpp ]
```
jeremy-murphy, line 117: Does it need to be a second file? Just wondering why.

jeremy-murphy (changes requested):
I haven't even got to the code proper yet but here are a few requests and questions to start with.

doc/maximum_weighted_matching.html
jorisvr: You are asking to remove `Verify` as parameter, and just never verify the result when called through this function. Right?

Note that verification requires access to the blossom structure and dual variables, which are not returned by this function. So a user who wants to verify, will have to write their own interface function that reaches into `boost::graph::detail::maximum_weighted_matching_context`.

At that point, should we consider to just remove the verification code entirely?
I'm fine with removing it. I personally think that always verifying has merits, but I agree it is a trade-off against complication of the code.

doc/maximum_weighted_matching.html
jorisvr: I see your point, but I'd like to understand how you want me to solve it.

The text already describes the benefit of each deviation. If you just want me to elaborate, I can e.g. expand each bullet into a full paragraph, provide some details, etc.

The reference paper has a focus on time complexity, not practical efficiency. From a complexity point of view, these deviations are trivial tinkering since the time complexity remains the same. If you want, I can provide specific arguments to point this out.

Reusing labels is not my original idea. It is widely used, for example Mehlhorn and Schäfer [1], LEMON, Kolmogorov [2]. But they all use different base algorithms, O(V E log(V)) and O(V^2 E) respectively, therefore the performance trade-off is different. I did some informal benchmarking on random graphs, but printing benchmarks in the documentation seems out-of-place to me.

[1] https://dl.acm.org/doi/abs/10.1145/944618.944622
[2] https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=930b9f9c3bc7acf3277dd2361076d40ff03774b2

test/Jamfile.v2
jorisvr: I was initially apprehensive about changing existing code and thought it best to keep the new stuff separate.

The files can be merged. The styles are quite different though. `weighted_matching_test` reads testcases from a data file, while `test2` builds testcases in commented C++ code. I prefer to keep the coded style for the new test cases.

Let me know if you want me to merge into a single test program.

doc/maximum_weighted_matching.html
jorisvr: Agreed. Fixed.

test/weighted_matching_test2.cpp
```diff
@@ -0,0 +1,522 @@
...
+//=======================================================================
+
+#include <boost/graph/maximum_weighted_matching.hpp>
+
+#include <iostream>
+#include <limits>
+#include <random>
+#include <sstream>
+#include <type_traits>
+#include <vector>
```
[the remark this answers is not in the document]
jorisvr: Fixed. Thanks.

doc/maximum_weighted_matching.html
jeremy-murphy: Yes, I think if we're not confident that the algorithm is reliable then we just have to test it more until we are. Releasing it with an option to check that the result is as expected just kind of screams of a lack of quality.
On principle, we should return these blooms and other things anyway as a tuple of values that the user can just ignore if they're not interested in them. That also makes it simple for us to do this verification when we need to, without having to modify the function. (Which will be the case any time we make a change to the logic.)
Now, having said that, I hope you don't have to make too many changes to the design just so that you can return these internals.

jeremy-murphy: I'm not 100% sure yet, but the verification code might not belong in the Boost.Graph library itself, but it would definitely belong in something that _fuzzes_ it. That does exist elsewhere, so that could be a next step after getting this PR finished.

jorisvr: Ok. It seems clear that we are not going to verify answers by default, and I agree with your reasoning for that decision.

I would then recommend to just remove the verification routine. Normal users are not going to use it. The verification logic itself is not trivial. Keeping such unused code in the library will just be a burden.

Putting the verification code in a fuzzer is interesting. I guess the fuzzer can just directly access the internal class to reach the data it needs. Alternatively, the fuzzer can compare answers against a trusted 3rd party implementation like LEMON.

Returning the blossom structure and duals is possible, but these data then need to be rearranged into a suitable structure at the end of the algorithm. During the algorithm, I need them in an efficient form. Returning this seems like more trouble than it is worth. It would also expose internal details about the algorithm, possibly complicating future revisions.

If you agree, I will add a commit to drop the verification stuff.

jeremy-murphy: Yeah, let's focus on the primary goal: fixing the implementation. We can always add more or tweak it later. So yes, for now let's remove the verification code altogether and don't worry about returning the blossom structures.

test/Jamfile.v2
jeremy-murphy: It's fine, leave them separate, but maybe rename their suffix to distinguish how they're different. It's not too important.

doc/maximum_weighted_matching.html
jorisvr: Done.

doc/maximum_weighted_matching.html
jorisvr: I expanded the text a bit to argue explicitly that each change preserves the overall time complexity and correctness of the algorithm.

jeremy-murphy: Thanks. I'm still not hugely comfortable with unpublished changes to a published algorithm, but I don't want to be rude and throw out your valuable contribution. 
Since after all it fixes the known bugs, I think the pragmatic thing to do is accept it as is and let the community figure out if it's right. :)

jeremy-murphy (changes requested):
Sorry, just a trivial request, but since the library is C++14 now can you please use `using` instead of `typedef` and remove spaces from in-between nested template <<....>> brackets? Thanks. I will enable clang-format one day...
PS. I mean, a template should be written as
`template <typename Foo>`, etc.

test/weighted_matching_test2.cpp
```diff
@@ -0,0 +1,522 @@
...
+        {4, 11, 933}, {5, 6, 608}, {5, 8, 636}, {6, 9, 274}, {6, 10, 279},
+        {7, 9, 705}, {7, 10, 114}, {8, 11, 602}, {9, 10, 764}, {10, 11, 413}},
+        4599);
+
+    // Random graphs:
+    std::mt19937 rng(123456);
+    for (int k = 0; k < 20000; ++k)
+        run_random_graph< long >(10, 15, rng);
+    for (int k = 0; k < 10000; ++k)
+        run_random_graph< double >(10, 15, rng);
```
jeremy-murphy, line 519: Please remove the random graphs, that should be moved to fuzz testing. Unit tests should specifically exercise different paths through the algorithm and edge cases, with as few tests as possible.

include/boost/graph/maximum_weighted_matching.hpp
```diff
@@ -49,1265 +63,1430 @@ matching_weight_sum(const Graph& g, MateMap mate)
     return matching_weight_sum(g, mate, get(vertex_index, g));
 }
 
+// brute-force matcher searches all possible combinations of matched edges to
+// get the maximum weighted matching which can be used for testing on small
+// graphs (within dozens vertices)
 template < typename Graph, typename MateMap, typename VertexIndexMap >
-class weighted_augmenting_path_finder
+class brute_force_matching
```
jeremy-murphy, line 70: I think we need to be specific about which algorithm this is, since everything is in the same namespace. Also, maybe _matcher_ is better than _matching_?

include/boost/graph/maximum_weighted_matching.hpp
```diff
@@ -49,1265 +63,1430 @@ matching_weight_sum(const Graph& g, MateMap mate)
...
 
-        // get all vertices inside recursively
-        virtual std::vector< vertex_descriptor_t > vertices() const
+        vertex_iterator_t vi, vi_end;
+        for (boost::tie(vi, vi_end) = vertices(g); vi != vi_end; ++vi)
+            put(pm, *vi, best_mate[*vi]);
+    }
+
+private:
+    const Graph& g;
```
jeremy-murphy, line 108: Add a general rule, please don't use references as member variables, as they prevent the class from being Regular. Uglier as it might be, just use a pointer.

include/boost/graph/maximum_weighted_matching.hpp
```diff
@@ -49,1265 +63,1430 @@ matching_weight_sum(const Graph& g, MateMap mate)
...
 
-        init();
+/** Check that vertex indices are unique and in range [0, V). */
+template < typename Graph, typename VertexIndexMap >
+void check_vertex_index_range(const Graph& g, VertexIndexMap vm)
+{
+    typedef typename property_traits< VertexIndexMap >::value_type index_t;
+    typedef typename std::make_unsigned<index_t>::type unsigned_index_t;
+    auto nv = num_vertices(g);
+    std::vector<bool> got_vertex(nv);
```
jeremy-murphy, line 173: Make sure this is really what you want, as opposed to a bitset.

include/boost/graph/maximum_weighted_matching.hpp
```diff
@@ -49,1265 +63,1430 @@ matching_weight_sum(const Graph& g, MateMap mate)
...
+    typedef std::pair< vertex_t, vertex_t > vertex_pair_t;
+
+    /**
+     * List of edges forming an alternating path or alternating cycle.
+     *
+     * The path is defined over top-level blossoms; it skips parts of the path
+     * that are internal to blossoms. Vertex pairs are oriented to match the
+     * direction of the path.
+     */
+    typedef std::deque< vertex_pair_t > alternating_path_t;
```
jeremy-murphy, line 219: deque is rarely faster than vector in practice: did you benchmark?

include/boost/graph/maximum_weighted_matching.hpp
```diff
@@ -49,1265 +63,1430 @@ matching_weight_sum(const Graph& g, MateMap mate)
...
+            /** Pointer to sub-blossom. */
+            blossom_t* blossom;
+
+            /** Vertex pair (x, y) where "x" is a vertex in "blossom" and
+                "y" is a vertex in the next sub-blossom. */
+            vertex_pair_t edge;
+        };
+
+        /** List of sub-blossoms, ordered along the alternating cycle. */
+        std::list< sub_blossom_t > subblossoms;
```
jeremy-murphy, line 301: And list is typically much much worse than vector; did you benchmark?

include/boost/graph/maximum_weighted_matching.hpp
```diff
@@ -49,1265 +63,1430 @@ matching_weight_sum(const Graph& g, MateMap mate)
...
-            && i->get() == j->get())
-        {
-            lowest_common_blossom = *i;
-            ++i;
-            ++j;
+    /** Call a function for every vertex inside the specified blossom. */
+    template < typename Func >
+    static void for_vertices_in_blossom(const blossom_t* blossom, Func func)
+    {
+        const nontrivial_blossom_t* ntb = blossom->nontrivial();
```
jeremy-murphy, line 479: Just use auto?

jeremy-murphy (commented):
Not quite finished...

test/weighted_matching_test2.cpp
jorisvr: Ok, done.

include/boost/graph/maximum_weighted_matching.hpp
jorisvr: I kept this piece of code unchanged from the previous version. But I'm happy to change it. Shall I make it `class brute_force_maximum_weighted_matcher` ?

include/boost/graph/maximum_weighted_matching.hpp
jorisvr: I believe `std::bitset<N>` requires its size to be fixed at compile time. But I need an array which is sized at run time to match the number of vertices of the graph.

include/boost/graph/maximum_weighted_matching.hpp
jorisvr: I add elements on both sides of the deque while tracing the path in two directions. To do that with a vector requires some extra tricks.

include/boost/graph/maximum_weighted_matching.hpp
jorisvr: I had not benchmarked, but I have now and there is no measurable difference. This list is not accessed very often.

I chose list because I use `splice()` to reorient the base of the list during augmentation. It is just a matter of convenience. I can instead use vector and either extra bookkeeping or copying of elements.

Do you prefer `std::vector` over `std::list` in cases where the performance is neutral?

include/boost/graph/maximum_weighted_matching.hpp
jorisvr: I don't understand.
Do you mean `auto func` instead of `template <typename Func>` ? That was not allowed before C++20.

Or do you mean `auto ntb` instead of `const nontrivial_blossom_t* ntb` ?

jeremy-murphy (commented):
Still going,..

jeremy-murphy:
There are some functions that take a `blossom*` but require that it is not null, in which case they should take a reference instead.

jeremy-murphy:
That's all from me, once the last comments are resolved I'll merge it in.
Overall I love the code, so thank you!

include/boost/graph/maximum_weighted_matching.hpp
jeremy-murphy: Yeah it's used in existing code, but it's a bad practice, so let's not add more.

jorisvr: Ok, done.

include/boost/graph/maximum_weighted_matching.hpp
jorisvr: Ok. I changed it to `auto` for these and a few similar verbose declarations.

Is that what you had in mind, or do you want to push further towards the *almost-alway-auto* style?

include/boost/graph/maximum_weighted_matching.hpp
```diff
@@ -49,1265 +63,1430 @@ matching_weight_sum(const Graph& g, MateMap mate)
...
+         * This is the unique vertex inside the blossom which is not matched
+         * to another vertex in the same blossom.
+         */
+        vertex_t base_vertex;
+
+        /** Label S or T or NONE, only valid for top-level blossoms. */
+        blossom_label_t label;
+
+        /** True if this is an instance of nontrivial_blossom. */
+        const bool is_nontrivial_blossom;
```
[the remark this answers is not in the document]
jorisvr: Done.

include/boost/graph/maximum_weighted_matching.hpp
```diff
@@ -10,1304 +11,1477 @@
...
-            vertex_descriptor_t min_tau_v
-                = graph_traits< Graph >::null_vertex();
-            edge_property_t min_tau
-                = std::numeric_limits< edge_property_t >::max();
-
-            std::vector< vertex_descriptor_t > sub_vertices
-                = sub_blossom->vertices();
-            for (vertex_vec_iter_t v = sub_vertices.begin();
-                 v != sub_vertices.end(); ++v)
+            if ((! edge.has_value()) || (s < slack))
```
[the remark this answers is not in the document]
jorisvr: Oh, I probably came up with this. I don't see it anywhere else in BGL.
I removed the spaces now.

I really like the space, but this is not the time to start a discussion about code formatting.

jorisvr:
> There are some functions that take a `blossom*` but require that it is not null, in which case they should take a reference instead.

I now changed these into `blossom&`.
This involved adding a number of `*` and `&` operators in the code. Because `blossom*` is still the native type in several data structures, I end up going back-and-forth between pointer and reference. In my opinion, it makes the code less clear. But I can live with it.

jeremy-murphy:
> > There are some functions that take a `blossom*` but require that it is not null, in which case they should take a reference instead.
> 
> I now changed these into `blossom&`. This involved adding a number of `*` and `&` operators in the code. Because `blossom*` is still the native type in several data structures, I end up going back-and-forth between pointer and reference. In my opinion, it makes the code less clear. But I can live with it.

I know what you mean, sometimes you just have to pay a price for doing things correctly, but it also might mean that those other places using `blossom*` could do with some refactoring.

include/boost/graph/maximum_weighted_matching.hpp
jeremy-murphy: I personally use and recommend AAA style, but what you've done here is fine. I just felt those long typenames with qualifiers were not helping the readability.

jeremy-murphy (approved):

include/boost/graph/maximum_weighted_matching.hpp
jeremy-murphy: Something recently reminded me that there is a better way to do this than using a raw pointer, which is to use `std::reference_wrapper`. 
https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper
It might not be a perfect solution but I would definitely have a go at using it to see if it improves readability.

jorisvr: > `std::reference_wrapper`

I did look into that, but I could not get it to work in a way that seems helpful.
In particular, the following does not work:

```
std::reference_wrapper<const Graph> g;

...
auto nv = num_vertices(g);
```

If I write that, I get a type error. Although `std::reference_wrapper` is supposedly implicitly convertable to the stored reference type, that doesn't seem to work in this case. Perhaps because `num_vertices()` is a function template, the reference wrapper itself ends up being passed as function argument.

I _can_ write `num_vertices(g.get())`
but I'm not sure it is much of an improvement over `num_vertices(*g)`.

include/boost/graph/maximum_weighted_matching.hpp
jeremy-murphy: I forgot to address this one, whoops. 
`vector<bool>` is generally avoided because of its unusual performance trying to satisfy the standard container interface (providing a reference to each element via iteration) whilst only using one bit per value, which requires the use of a proxy class, etc. 
A dynamic bitset is just more honest about what it is and does.
It's not a big deal, might not need to be changed until someone makes some other changes in there.

include/boost/graph/maximum_weighted_matching.hpp
jeremy-murphy: Ahhh, yeah, it doesn't magically remove in that case. Oh well, too bad.

PR #403, opened 2024-12-12
Make adjacency_matrix implement BidirectionalGraph concept

### LINKED ISSUE #404: `adjacency_matrix` should implement BidirectionalGraphConcept
I think there's no reason not to do this. Indeed, one can see that the majority of functions, required for the concept is implemented in a dedicated section. Please see https://github.com/boostorg/graph/pull/403 that finalizes that idea.

Wavetrace:
Adjacency matrix is by nature bidirectional and one can see that the concept was mostly implemented before. That implementation missed the degree function(s). These functions are implemented in this commit and the traversal_category is updated to meet the concept requirements.

Tests are also added, though they are a bit shallow in the sense that they seem to omit intensive algorithms.

This PR is a consequence of work on https://github.com/boostorg/graph/pull/397 -- that one contains a test commented out, that ends up running `neighbor_bfs` on an adjacency matrix.

Fixes #404

### DIFF
--- a/include/boost/graph/adjacency_matrix.hpp
+++ b/include/boost/graph/adjacency_matrix.hpp
@@ -433,7 +433,8 @@ struct adj_matrix_traversal_tag : public virtual adjacency_matrix_tag,
                                   public virtual vertex_list_graph_tag,
                                   public virtual incidence_graph_tag,
                                   public virtual adjacency_graph_tag,
-                                  public virtual edge_list_graph_tag
+                                  public virtual edge_list_graph_tag,
+                                  public virtual bidirectional_graph_tag
 {
 };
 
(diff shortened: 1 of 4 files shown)


include/boost/graph/adjacency_matrix.hpp
```diff
@@ -826,6 +827,26 @@ typename adjacency_matrix< D, VP, EP, GP, A >::degree_size_type in_degree(
     return n;
 }
 
+// degree
```
jeremy-murphy: This comment is not doing much work. :)

jeremy-murphy (changes requested):

jeremy-murphy:
I thought it would be expedient to just fix the comments myself since the changes were trivial.

jeremy-murphy (approved):
Thanks for a pragmatic missing piece of the puzzle!

jonasliljenfeldt:
Nice one @Wavetrace!

Wavetrace:
Thanks, Jonas! Thanks a lot, Jeremy, for fixing the comment!

PR #409, opened 2024-12-21
Replace Boost::foreach dependency with range-based for

georgthegreat:
The library is marked as C++14, hence it should be safe to replace the dependency with native syntax.

`boost::graph::vertices` returns `std::pair` which is not compatible with range-based for.
As changing this behavior would result in API break, only public part was changed and the dependency still remains.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,6 @@ target_link_libraries(boost_graph
     Boost::conversion
     Boost::core
     Boost::detail
-    Boost::foreach
     Boost::function
     Boost::integer
     Boost::iterator
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,7 @@ target_link_libraries(boost_graph
     Boost::conversion
     Boost::core
     Boost::detail
+    Boost::foreach
     Boost::function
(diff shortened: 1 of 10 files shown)


jeremy-murphy:
Thanks!

jeremy-murphy:
If there are any build failures, please investigate.

georgthegreat:
@jeremy-murphy, I had to reduce the amount of changes as it turns out that range-based-for is incompatible with vertices() method (it returns std::pair which is not a range in c++ terms, yet is accepted by BOOST_FOREACH).

So, publicpart is rewritten and boost/foreach is no longer included from neither `include/` nor `src/` directories.

Can we merge such (partial) improvement?

jeremy-murphy:
Whilst I think it's good to get rid of macros like `BOOST_FOREACH`, what I think is even better is to use algorithms in preference to `for ()` loops. And it turns out that the implementation in Boost.Ranges _does_ support `boost::pair` (and I _presume_ `std::pair`) as a range. However, I don't want you to go overboard with refactoring right away, as many of these loops have large bodies, so I recommend picking the low-hanging fruit, the stuff that looks obviously like a `find_if`, etc.

PR #410, opened 2024-12-25
Fixed wrong link in grid_graph documentation

Tuxliri:
There was an erroneous link in the `grid_graph` documentation.


jeremy-murphy (approved):
Nice catch, thanks!

PR #412, opened 2024-12-27
Fix geometry bug in is_straight_line_drawing

jeremy-murphy:
Fix the geometry bug uncovered and documented by @Hermann-SW by replacing the custom function with one from Boost.Geometry.
`crosses` meets the requirement described in the old code for returning true for 'intersections' that don't include endpoints or self-intersection.
I decided to remove the tolerance altogether on the assumption that it was indicative of a poor implementation.

### DIFF
--- a/include/boost/graph/is_straight_line_drawing.hpp
+++ b/include/boost/graph/is_straight_line_drawing.hpp
@@ -19,6 +19,8 @@
 #include <boost/geometry/algorithms/crosses.hpp>
 #include <boost/geometry/geometries/linestring.hpp>
 
+#include <boost/numeric/conversion/cast.hpp>
+
 #include <algorithm>
 #include <vector>
 #include <map>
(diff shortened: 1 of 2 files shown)


jeremy-murphy:
> Why is this assigned to me?

Sorry I meant to leave a comment that I just wanted you to give it a sanity check. I've never used the numeric_cast before, but I presume it's as simple as this.

jeremy-murphy:
I just don't really know who else to assign my own pull requests to, since I assume JZ Maddock wants to minimise his Boost workload these days.

jeremy-murphy:
Whoops, accidentally pushed my CMake changes here and force-pushed to get rid of them.

Becheler (approved):
The implementation aligns well with the general style and structure of the Boost Graph Library.

The introduction of a new dependency on Boost::Geometry was briefly discussed in issue #388 and deemed worthwhile during that discussion.

The function is well-organized but remains somewhat lengthy despite the effective code reduction by @jeremy-murphy. While I recognize that adding further function extraction, comments, or documentation could help new contributors better understand the context and usage, I understand that this is not the primary focus of this quick fix.

The code appears correct, and all tests pass; however, I am not deeply familiar with planar graph theory. It might be beneficial for someone with more expertise in this area to provide additional input, though @jeremy-murphy’s expertise may already suffice.

In summary, the tests pass, and the code is consistent with the library's conventions. While I may not be fully qualified to evaluate the deeper theoretical aspects, I’m happy to assist further if needed!

PR #416, opened 2025-02-05
Remove Boost.Assign usages.

gogagum:
Boost.Assign is not needed C++14 code. This PR removes this library usages. It is also suggested to use `std::vector` `std::initializer_list` constructor instead of series of `push_back` calls in test/two_graphs_common_spanning_trees_test.cpp.


jeremy-murphy (approved):
Thanks for a tidy improvement to the code! Just waiting for the tests to run.

PR #418, opened 2025-02-11
Remove Boost.Bind usages.

gogagum:
This PR replaces `boost::bind` usages with lambdas, so **Boost.Bind** can be removed from dependances.

### DIFF
--- a/include/boost/graph/king_ordering.hpp
+++ b/include/boost/graph/king_ordering.hpp
@@ -48,9 +48,6 @@ namespace detail
         template < typename Vertex, typename Graph >
         void finish_vertex(Vertex, Graph& g)
         {
-            using std::placeholders::_1;
-            using std::placeholders::_2;
-
             typename graph_traits< Graph >::out_edge_iterator ei, ei_end;
             Vertex v, w;
 
@@ -61,7 +58,7 @@ namespace detail
             reverse_iterator rbegin = Qptr->rbegin();
 
             // heap the vertices already there
-            std::make_heap(rbegin, rend, std::bind(comp, _2, _1));
+            std::make_heap(rbegin, rend, [this](const auto& first, const auto& second) { return comp(second, first); });
 
             unsigned i = 0;
 
--- a/include/boost/graph/king_ordering.hpp
+++ b/include/boost/graph/king_ordering.hpp
@@ -15,7 +15,6 @@
 #include <vector>
 #include <algorithm>
 #include <boost/config.hpp>
-#include <boost/bind/bind.hpp>
 #include <boost/tuple/tuple.hpp>
 #include <boost/graph/detail/sparse_ordering.hpp>
 #include <boost/graph/graph_utility.hpp>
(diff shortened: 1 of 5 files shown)


gogagum:
I will make an attempt of doing some workaronds for Clang.

jeremy-murphy:
Thanks for your effort to improve Boost.Graph, it's really appreciated! 
However, I strongly recommend you watch this video about `bind` and then consider what change you want to make: https://youtu.be/zt7ThwVfap0?si=47AbaJkWFh-VrnBc

gogagum:
Well, the speaker recommends to avoid `std::bind`. Do you mean this? I just tried to replace `boost::bind` with `std::bind` where possible to make minimal difference. It is not a problem to replace all `std::bind` usages with lambdas.

jeremy-murphy:
> Well, the speaker recommends to avoid `std::bind`. Do you mean this? I just tried to replace `boost::bind` with `std::bind` where possible to make minimal difference. It is not a problem to replace all `std::bind` usages with lambdas.

Yeah, I think replacing one `bind` with another is not really worth doing when it could be replaced with something even better.

include/boost/graph/king_ordering.hpp
```diff
@@ -14,8 +14,8 @@
 #include <deque>
 #include <vector>
 #include <algorithm>
+#include <functional>
```
jeremy-murphy: Still needed?

include/boost/graph/transitive_closure.hpp
```diff
@@ -130,16 +129,16 @@ void transitive_closure(const Graph& g, GraphTC& tc,
...
         std::pair< cg_adj_iter, cg_adj_iter > pr = adjacent_vertices(i, CG);
         CG_vec[i].assign(pr.first, pr.second);
         std::sort(CG_vec[i].begin(), CG_vec[i].end(),
-            boost::bind(std::less< cg_vertex >(),
-                boost::bind(detail::subscript(topo_number), _1),
-                boost::bind(detail::subscript(topo_number), _2)));
+            [&topo_number](const auto& cg_0, const auto& cg_1)
+            {
+                return std::less< cg_vertex >()(
+                    topo_number[cg_0], topo_number[cg_1]);
```
jeremy-murphy, line 140: Can we just use `a < b` now that it's in a lambda?

jeremy-murphy (changes requested):

jeremy-murphy:
I requested changes, but really they're just questions/suggestions. This code is good to go if there is a reason for everything.

include/boost/graph/transitive_closure.hpp
gogagum: I did not catch that `cg_vertex` if just a `size_t`, so my idea was to suggest the closest variant without analyzing what `cg_vertex` is. I would agree that it can be simplified, but there is a good thing about `std::less`: it checks the type of its arguments. So if the type for topo number is changed, there will be an compile error.

jeremy-murphy: OK, cool, I'm happy to leave it as is.

jeremy-murphy (approved):

jeremy-murphy:
Great, thanks for this improvement!

PR #421, opened 2025-02-16
examples: C++11: Some use of range-based for loops

murraycu:


example/topo-sort1.cpp
```diff
@@ -35,9 +35,8 @@ int main()
     topological_sort(g, std::front_inserter(topo_order),
         vertex_index_map(identity_property_map()));
 
-    int n = 1;
-    for (auto i = topo_order.begin(); i != topo_order.end(); ++i, ++n)
-        std::cout << tasks[*i] << std::endl;
```
jeremy-murphy, line 33: `n` was not even used?

jeremy-murphy (approved):

example/cuthill_mckee_ordering.cpp
```diff
@@ -57,8 +57,8 @@ int main(int, char*[])
         Pair(6, 7) }; // g-h
 
     Graph G(10);
-    for (int i = 0; i < 14; ++i)
-        add_edge(edges[i].first, edges[i].second, G);
+    for (const auto& edge : edges)
```
jeremy-murphy: Sorry, you made one west const here.

jeremy-murphy (changes requested):

example/cuthill_mckee_ordering.cpp
murraycu: Thanks. Done.

example/topo-sort1.cpp
murraycu: Apparently not.

jeremy-murphy (approved):
Great, thanks for another modernization!

PR #422, opened 2025-02-16
examples: C++11: Use list initialization instead of push_back()

murraycu:


jeremy-murphy (approved):
Thanks for another easy-to-review improvement!

PR #424, opened 2025-02-16
examples: C++11: Use std::begin() and std::end()

murraycu:
Instead of array + length.
And use initializer lists.

### DIFF
--- a/example/bellman-ford-internet.cpp
+++ b/example/bellman-ford-internet.cpp
@@ -25,19 +25,18 @@ int main()
         H,
         n_vertices
     };
-    const int n_edges = 11;
     typedef std::pair< int, int > Edge;
 
     // The list of connections between routers stored in an array.
-    Edge edges[] = { Edge(A, B), Edge(A, C), Edge(B, D), Edge(B, E), Edge(C, E),
-        Edge(C, F), Edge(D, H), Edge(D, E), Edge(E, H), Edge(F, G),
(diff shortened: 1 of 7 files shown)


example/bellman-ford-internet.cpp
```diff
@@ -29,15 +29,15 @@ int main()
...
     // The list of connections between routers stored in an array.
-    Edge edges[] = { Edge(A, B), Edge(A, C), Edge(B, D), Edge(B, E), Edge(C, E),
-        Edge(C, F), Edge(D, H), Edge(D, E), Edge(E, H), Edge(F, G),
+    const auto edges = { Edge(A, B), Edge(A, C), Edge(B, D), Edge(B, E),
+        Edge(C, E), Edge(C, F), Edge(D, H), Edge(D, E), Edge(E, H), Edge(F, G),
         Edge(G, H) };
 
     // Specify the graph type and declare a graph object
-    typedef edge_list< Edge*, Edge, std::ptrdiff_t,
+    typedef edge_list< const Edge*, Edge, std::ptrdiff_t,
```
jeremy-murphy: I'm not sure that constant data is a good way to demonstrate the library, just because being able to change edges and mutate graphs is such a normal practice. If users learn to create const data to start with, I expect they'll get frustrated as they try to do things with it until they eventually realise that the const was what was holding them back. 
This is very much an argument based on psychology rather than technical merit or const-correctness. :)

murraycu: OK. Done.

jeremy-murphy (approved):

PR #423, opened 2025-02-16
examples: C++11: Use using instead of typedef

murraycu:

PR #425, opened 2025-03-15
examples: C++11: Use using instead of typedef.

murraycu:


murraycu:
Sorry, I seem to have proposed this already in https://github.com/boostorg/graph/pull/423.

PR #429, opened 2025-05-01
Fix multiplicity value comparison

### LINKED ISSUE #428: boost::isomorphism runs into an infinite loop
## How to reproduce

* compile this code:

```
   struct VertexProps
   {
        std::string node_id; // will store "0", "1", ..., "74"
   };

using Graph = boost::adjacency_list<boost::vecS, boost::vecS, boost::bidirectionalS, VertexProps>;
void  loadGraphFromDOT(const std::string& filename, Graph& g)
{
   std::ifstream in(filename);
   if (!in)
   {
        throw std::runtime_error("Error: Cannot open file ");
        
   }
   auto node_id_map = boost::get(&VertexProps::node_id, g);
   boost::dynamic_properties dp;
   dp.property("node_id", node_id_map);

   if (!boost::read_graphviz(in, g, dp))
   {
        throw std::runtime_error("Error: Failed to read DOT ");        
   }
}



TEST(GraphTest, isomerphismTest)
{
   using Graph = boost::adjacency_list<boost::vecS, boost::vecS, boost::bidirectionalS, VertexProps>;
   Graph g0;
   Graph g1;
   loadGraphFromDOT("graph0.dot", g0);
   loadGraphFromDOT("graph2.dot", g1);

   const bool iso = boost::isomorphism(g0, g1);
   EXPECT_TRUE(iso);
}
```

* run the code with the given dot files  (adjust the file ending)
[graph0.dot.txt](https://github.com/user-attachments/files/19773816/graph0.dot.txt)
[graph2.dot.txt](https://github.com/user-attachments/files/19773817/graph2.dot.txt)

*  boost::isomorphism will never finish
* I would expect 'true' as a result for the function
##  Version and compiler infos

* I encountered the issue while upgrading boost's version from 1.81.0 to 1.86.0. It worked fine with 1.81.0 and does not work with 1.86.0.
* compiler version is  ````Microsoft (R) C/C++ Optimizing Compiler Version 19.38.33133 for x86````

jan-grimo:
Replacement of at() in 8f3139be as part of #349 was faulty: Iterator deref is a (key, value) pair, but only the value should be compared

Closes #428 

- Fixes bug in `compare_multiplicity::operator bool`
- Adds test case from #428 into isomorphism test set


jan-grimo:
@jeremy-murphy can you help me out with the failing CI tests? I'm not sure what's wrong, perhaps it's the use of the graphviz reader? Is there a better way I should be embedding the sample graphs from the reporting issue?

jeremy-murphy:
Yes, there is a problem with the CI that I'll fix today.

jeremy-murphy:
Looks like there might also actually be a problem with file access in your new test on macOS?
```
====== BEGIN OUTPUT ======
Elapsed time (clock cycles): 1050
Elapsed time (clock cycles): 1776
libc++abi: terminating due to uncaught exception of type std::runtime_error: Error: Cannot open file 

EXIT STATUS: 134
====== END OUTPUT ======
```

jeremy-murphy:
The CI is fixed now, so please merge the tip of `develop` in and you should get a green build.

jeremy-murphy:
Looks like a problem with your change?

jeremy-murphy:
@jan-grimo ping.

jeremy-murphy:
Thanks for fixing it promptly!

PR #432, opened 2025-07-10
Update r_c_shortest_paths.html

andrea-cassioli-maersk:
Add few details about the on_enter_loop visitor event that is not documented.

### DIFF
--- a/doc/r_c_shortest_paths.html
+++ b/doc/r_c_shortest_paths.html
@@ -329,6 +329,11 @@ If <tt>vis</tt> is an object of a type modelling the ResourceConstrainedShortest
 <table border>
 <tr>
 <td>
+<tt>vis.on_enter_loop(unprocessed_labels, const Graph&amp; g )</tt>
+</td>
+</tr>
+<tr>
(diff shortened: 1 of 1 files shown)


doc/r_c_shortest_paths.html
```diff
@@ -329,6 +329,11 @@ <h4>DominanceFunction</h4>
 <table border>
 <tr>
 <td>
+<tt>vis.on_enter_loop(unprocessed_labels, const Graph&amp; g )</tt>
```
jeremy-murphy: Shouldn't it rather be this?
```suggestion
<tt>vis.on_enter_loop(const Queue&amp; queue, const Graph&amp; g )</tt>
```
It looks like you accidentally mixed arguments and parameters together, but I could be wrong.

andrea-cassioli-maersk: You are right @jeremy-murphy , I was unsure about the convention in the docs and which type to use.

jeremy-murphy:
Thank you!

Issue #437, opened 2025-08-21
sprintf deprecated


sdarwin:
"Use snprintf if available as some compilers (clang 14.0) issue deprecation warnings for sprintf"

```
libs/graph/example/iohb.c:    sprintf(iformat,"%%%dd",Indwidth);
libs/graph/example/iohb.c:    sprintf(iformat,"%%%dd",Indwidth);
libs/graph/example/iohb.c:    sprintf(pformat,"%%%dd",Ptrwidth);
libs/graph/example/iohb.c:    sprintf(pformat,"%%%dd",Ptrwidth);
libs/graph/example/iohb.c:          sprintf(rformat,"%% %d.%dE",Rhswidth,Rhsprec);
libs/graph/example/iohb.c:          sprintf(rformat,"%% %d.%df",Rhswidth,Rhsprec);
libs/graph/example/iohb.c:       sprintf(rformat,"%%%ds",Rhswidth);
libs/graph/example/iohb.c:         sprintf(vformat,"%% %d.%dE",Valwidth,Valprec);
libs/graph/example/iohb.c:         sprintf(vformat,"%% %d.%df",Valwidth,Valprec);
libs/graph/example/iohb.c:      sprintf(vformat,"%%%ds",Valwidth);
```

jeremy-murphy:
I just merged the fix for that one but I'll let you close it when you confirm that there are no other occurrences that need fixing.

sdarwin:
" I'll let you close it"...  Alright. Thanks. Those were the occurrences of `sprintf`.

PR #438, opened 2025-08-24
Update iohb.c to use snprintf

Becheler:
Use snprintf if available as some compilers (clang 14.0) issue deprecation warnings for sprintf, see issue #437


jeremy-murphy:
We have an example in C???

jeremy-murphy:
Although some of those checks failed/cancelled for weird reasons, the important checks passed.

PR #441, opened 2025-10-13
Add Boost.Geometry to dependencies

jeremy-murphy:


jeremy-murphy:
@grafikrobot , did I add it correctly?

Or is BGL the first library to depend on Boost.Geometry and test this part of the build system?

jeremy-murphy:
```
  The link interface of target "boost_geometry" contains:

    Boost::crc

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

```
OK, so this is a bug in Boost.Geometry?

jeremy-murphy:
@pdimov do you maintain Boost.CRC? Do you know if this problem is with Geometry or CRC?

pdimov:
It's not a problem in CRC. It's caused by Geometry listing CRC as a dependency in its CMakeLists.txt here:

https://github.com/boostorg/geometry/blob/8aaf504091cfb46541b0a014cb0778c4d7754b51/CMakeLists.txt#L27

but it doesn't actually depend on CRC.

jeremy-murphy:
@pdimov thanks for answering. But... why does that cause this error, just because it doesn't actually use it?

pdimov:
Because only the actual dependencies are installed by depinst, so libs/crc is empty and the target Boost::crc doesn't exist.

jeremy-murphy:
Thanks for the explanation. PR to fix Boost.Geometry: https://github.com/boostorg/geometry/pull/1437

jeremy-murphy:
Made redundant by new implementation that doesn't use Boost.Geometry.

PR #443, opened 2025-12-05
Fix adjacency list serialise types

andreacassioli:
This fixes #291.

According to [the offical docs](https://www.boost.org/doc/libs/latest/libs/graph/doc/graph_concepts.html) the type representing the number of vertices and edges in a graph can be deduced by the graph traits.

In this PR 

* read/write data types deduced from the graph traits
* small cleanup modernizing a bit the code

### DIFF
--- a/include/boost/graph/adj_list_serialize.hpp
+++ b/include/boost/graph/adj_list_serialize.hpp
@@ -45,8 +45,7 @@ namespace serialization
     {
         typedef adjacency_list< OEL, VL, D, VP, EP, GP, EL > Graph;
         typedef typename graph_traits< Graph >::vertex_descriptor Vertex;
-
-        typedef typename boost::graph_traits<Graph>::vertices_size_type VerticesSize;
+        typedef typename graph_traits< Graph >::vertices_size_type VerticesSize;
 
         VerticesSize V = num_vertices(graph);
         VerticesSize E = num_edges(graph);
@@ -86,10 +85,11 @@ namespace serialization
         typedef adjacency_list< OEL, VL, D, VP, EP, GP, EL > Graph;
         typedef typename graph_traits< Graph >::vertex_descriptor Vertex;
         typedef typename graph_traits< Graph >::edge_descriptor Edge;
+        typedef typename graph_traits< Graph >::vertices_size_type VerticesSize;
 
-        unsigned int V;
+        VerticesSize V;
         ar >> BOOST_SERIALIZATION_NVP(V);
-        unsigned int E;
+        VerticesSize E;
         ar >> BOOST_SERIALIZATION_NVP(E);
 
         std::vector< Vertex > verts(V);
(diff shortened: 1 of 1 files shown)


andreacassioli:
I think this might break compatibility because the size of the integers written and read will depends on the graphs used to read and write. So I believe there is a bit more to do.

andreacassioli:
@jeremy-murphy there is a failure in the CI that I do not understand, it does seem to be related to my changes. What do you think?

jeremy-murphy:
> @jeremy-murphy there is a failure in the CI that I do not understand, it does seem to be related to my changes. What do you think?

Yeah, that's very odd. Looks like a problem either in Boost.Math or in the system libraries themselves. I would just ignore it for now and hope it goes away by itself.

include/boost/graph/adj_list_serialize.hpp
```diff
@@ -43,17 +43,18 @@ namespace serialization
         const unsigned int /* file_version */
     )
     {
-        typedef adjacency_list< OEL, VL, D, VP, EP, GP, EL > Graph;
-        typedef typename graph_traits< Graph >::vertex_descriptor Vertex;
+        using Graph = adjacency_list< OEL, VL, D, VP, EP, GP, EL >;
+        using Vertex = typename graph_traits< Graph >::vertex_descriptor;
+        using SerializedInteger = unsigned int;
```
jeremy-murphy: Isn't this going to cause a problem for users with, however unlikely, graphs where `vertices_size_type` is `std::size_t` and actually have billions or trillions or vertices?

Also, it would waste space for users that have billions of tiny graphs where they have customized `vertices_size_type` to be `unsigned char`, etc?

andreacassioli: yeah, I follow the current implementation approach of storing always the same type (i.e. int) so that one can read/write with different graph type. It is indeed a bit wasteful. 

Otherwise one could deduce these types fro the traits, assuming that the graph type remains the same. It was actually my first thought.

andreacassioli: I gave it another shot. Now I assume types are deduced from the graph.

andreacassioli: @jeremy-murphy what you think?

include/boost/graph/adj_list_serialize.hpp
```diff
@@ -43,17 +45,17 @@ namespace serialization
...
+        const auto V = num_vertices(graph);
+        const auto E = num_edges(graph);
         ar << BOOST_SERIALIZATION_NVP(V);
         ar << BOOST_SERIALIZATION_NVP(E);
 
         // assign indices to vertices
-        std::map< Vertex, int > indices;
-        int num = 0;
+        std::unordered_map< Vertex, Vertex > indices(V);
+        Vertex num = 0;
```
jeremy-murphy, line 58: This is probably fine, because I can't imagine a vertex descriptor that is not some kind of integer, but occasionally we have to be careful about changing the requirements on a concept from, in this case, ordered by <, to, hashable.
Now, `std::unordered_map` is dreadfully slow, so please try `boost::unordered_flat_map`, which I assume will work fine. Once you've made that change, this PR is ready to be merged.

jeremy-murphy (changes requested):

include/boost/graph/adj_list_serialize.hpp
jeremy-murphy: I decided to merge it anyway. 😅

PR #449, opened 2026-01-15
Rcsp cleanup

### LINKED ISSUE #342: Is EdgeIndexMap needed in r_c_shortest_paths ?
I have been using `r_c_shortest_paths` for a while, and recently because of a code refactoring I realised that one of the template parameters, namely `EdgeIndexMap`, seems not to be used anywhere in `r_c_shortest_paths`. I was then wonder whether I am missing its purpose. Could anyone shed a light?

andreacassioli:
In this PR I introduce a breaking change to the RCSP algorithm interface along several modernization of the code

Breaking change:
* the RCSP interface does NOT require an edge map (it never had) and thus it should not be required. Thus I remove it all along. This fixes #342 

Other changes:
* the edge feasibility flag is moved to be local in to the extension function evaluation
* when accessing the final node list at destination, use a reference instead of a copy
* when a single solution is asked for, move it instead of copy.
* removed macro selecting the allocator deponding on C++ version is obsolete because it tests C++11 which is not supported.
* docs updated

Modernization
 
* the code is simplified by using `auto` extensively, that allows fewer type definitions and less verbose code
* some `typedef` has been rewritten with `using`
* use `std::tie` instead of `boost::tie`


jeremy-murphy:
Great, thanks, I'll try to have a close look soon.

Becheler:
HI @jeremy-murphy !
Just pinging you here, as this PR is interface-breaking it will require your supervision.
I had a glance over the PR and it seems good. Maybe some aggressive `auto`-ing but without proper guidelines from us it's expected 😄 

It should also be merged before #491 since it touches documentation page removed by #491

andreacassioli:
@Becheler I think this PR shall not be a blocker for https://github.com/boostorg/graph/pull/491 - I can update the new docs after the merge if it fits better.

I am curious about the guidelines for `auto`, maybe some more to consider in the contribution guidelines?

Becheler:
@andreacassioli the merge order is mostly to simplify your life so you don't have to rebase again + solve conflicts once the file you edited got deleted 😄 

As for the `auto` guidelines, yes I'm wondering about what those could be. My current mental model is that `auto` optimises for the writer's convenience at the cost of the reader's understanding. A new contributor should understand a variable's type, semantic and purpose without running a compiler. I guess something around those lines:

Use `auto` for: 

- Range-based for loops: `for (auto v : boost::make_iterator_range(vertices(g)))`
- Lambdas: `auto fun = [](){ ... }`
- Iterator variables: `auto it = container.begin()`
- Structured bindings (when/if C++17): `auto [ei, ei_end] = edges(g)`

Do not use `auto` :

- When a meaningful type alias already exists (use the alias)
- Function return types in headers: forces readers to mentally execute the function
- Anywhere a narrowing conversion could hide: `size_t`/`int` mismatches should be visible, not obscured
- Variables where the type carries semantic meaning : spell out `vertex_descriptor`, `edge_descriptor`, `size_type`
- Property map expressions : `auto x = get(vertex_index, g)` hides what kind of map `x` is. Use a type alias or explicit type

I think in highly generic library code like BGL we will tend to desire explicitness for review more than type hiding.

But this is open discussion at this point, what do you think? I haven't talked about it with Jeremy - I will raise the point next time we chat :)

andreacassioli:
@Becheler I have a bit of a different habit on `auto`: basically use it always except

* if a cast of sort is needed
* if the type shall be used after somehow (like for traits)
* if readibility is a problem (like call to a library that use some unclear naming)

that's pretty much it. I actually think that in a generic setting it really fulfill its purpose, i.e. when the logic should abstract the type. Also cumbersome long typedef to extract hidden types are not readable at all to me.

But again, personal taste. And it is a topic worth an entire discussion thread!

For this PR, shall I take a round to see if I have push a bit far? or shall I wait for the review?

Becheler:
Hi @andreacassioli 🥳

Yes it's an entire discussion thread to have to reach a proper guideline. I will try to summarize my take, your take and Jeremy's in a distinct RFC. From what I understand Jeremy leans more towards your philosophy 🤓

Regardless of the auto thing, I think it should not prevent this PR from being merged: you cleaned the code, removed 4 warnings on all matrix (congrats 🥳🤩) and the tests pass and the code base ends better. I will have a closer look tomorrow on the details - I got stuck today trying to rerun this stubborn ci drone and was putting your PR in the red 😛

Good night up there in the North !

include/boost/graph/r_c_shortest_paths.hpp
```diff
@@ -498,45 +479,44 @@ namespace detail
...
     {
     }
     template < class Label, class Graph >
-    void on_label_not_dominated(const Label&, const Graph&)
+    void on_label_not_dominated(const Label&, const Graph&) const 
     {
     }
     template < class Queue, class Graph >
-    bool on_enter_loop(const Queue& queue, const Graph& graph)
+    bool on_enter_loop(const Queue&, const Graph&) const
```
Becheler: I'd drop the const on the default visitor's hooks. The default visitor is the example people copy, and const here implies hooks must be observers, but the algorithm calls them on a non-const by-value Visitor, and stateful visitors (counting labels, recording the best path, logging) are a normal use case whose hooks must mutate. It's also inconsistent with the rest of BGL (e.g. default_dfs_visitor hooks are non-const), and since these are no-ops it adds no value, just a misleading signal. If the intent were to enforce observer-only hooks, that would require taking the visitor by const& in the dispatch, which would be a breaking change and would forbid stateful visitors, so I don't think we want that.

Becheler (changes requested):
thank you andrea !
Looks good although I would refrain from modifying the visitor public API - that would deviate from existing practices.

Becheler:
@jeremy-murphy this looks like an API deprecation, what should we do ? Expose the new signature will keeping the old one but marking it deprecated ?

include/boost/graph/r_c_shortest_paths.hpp
andreacassioli: You are right, reverted.

jeremy-murphy:
I understand that it's difficult to draw a line when tidying up old code: fixing one thing leads to discovery of another ten things, etc, but isolation of changes is essential to smooth operation. If we're proposing a breaking change, then I want to see only the changes that are absolutely essential to it. Otherwise I'm just swimming in unrelated code changes and I can't make a decision. It is much better to open 7 small pull requests than one pull request with 7 thematically related changes. Also think of it this way, do you want the importance of a breaking API change to rest on the acceptance of an extra `const` somewhere? They're two very different decisions and so they belong in different PRs.

include/boost/graph/r_c_shortest_paths.hpp
```diff
@@ -163,68 +163,59 @@ namespace detail
...
-        const EdgeIndexMap& /*edge_index_map*/,
         typename graph_traits< Graph >::vertex_descriptor s,
         typename graph_traits< Graph >::vertex_descriptor t,
         // each inner vector corresponds to a pareto-optimal path
         std::vector<
             std::vector< typename graph_traits< Graph >::edge_descriptor > >&
             pareto_optimal_solutions,
         std::vector< Resource_Container >& pareto_optimal_resource_containers,
-        bool b_all_pareto_optimal_solutions,
+        const bool b_all_pareto_optimal_solutions,
```
jeremy-murphy, line 178: It might seem counter-intuitive, but we never make _values_ in parameters `const`. Only references, or values pointed to.

Becheler:
Agreed, and a bit more granularity in the PRs would help, though as you said it does trade off slightly against contributor ease of life (more branches/PRs to track).

One thing I noticed in #496 : because the general cleanup wasn't kept separate from the API deprecation, @andreacassioli was *this* close to get us a perfectly clean clang CI for 1.92 by clearing the 4 remaining warnings 🥳  :

image


Saving that satisfaction for 1.93 🚀 
Thanks again @andreacassioli 😉

andreacassioli:
@jeremy-murphy and @Becheler , I agree that I could have split the PR in few smaller ones. But I could not resist and I thought most were fairly simple changes. But let me know if you want me to split or go ahead with this

jeremy-murphy:
I'll try to have another look tomorrow morning.

include/boost/graph/r_c_shortest_paths.hpp
```diff
@@ -25,7 +25,7 @@ namespace boost
 // r_c_shortest_paths_label struct
 template < class Graph, class Resource_Container >
 struct r_c_shortest_paths_label
-: public boost::enable_shared_from_this<
+: boost::enable_shared_from_this<
```
jeremy-murphy, line 28: Why?

jeremy-murphy:
> @jeremy-murphy and @Becheler , I agree that I could have split the PR in few smaller ones. But I could not resist and I thought most were fairly simple changes. But let me know if you want me to split or go ahead with this

Yeah, sorry, there are too many small things that aren't quite right. A PR to make a breaking change should not include anything extra.

andreacassioli:
OK, very fair, I understand. I then close this PR and soon open a new focused on the breaking change.

All learning for me! Thanks @jeremy-murphy and @Becheler for taking the time!

jeremy-murphy:
Thanks, I really appreciate your work, @andreacassioli .

Becheler:
Thank you @andreacassioli and sorry 😣 

If you split this in 2 PRs, I think the deprecation one should 

1) add your newly cleaned overload 

2) mark the old api as deprecated for 2 releases  
```cpp
#include <boost/config.hpp>

BOOST_DEPRECATED("Use new_algorithm() instead; removal planned for Boost 1.95")
void old_algorithm(Graph& g);
```

So:
- 1.93 will be the deprecation release 
- 1.94 will repeat the warning 
- 1.95 will effectively remove the old signature

Deprecation cycles are a bit of a hassle in Boost, mostly because so many projects in the wild depend on it and we have no way to figure out which projects and how to warn them directly. I suspect this is why things kept being added to the project, with little to no of the important cleanup you're doing. ♥️

To be transparent Jeremy and I will be in the same conundrum when we will mark all named parameters interfaces deprecated 😖

andreacassioli:
@Becheler how do you manage examples and tests in the first PR?

Becheler:
@andreacassioli I would say that since the change should be non breaking, and the PR should be minimal, you should probably:
- let the examples use the old API for now (so PR noise is reduced - we can update examples later)
- don't remove tests of the old API, since it will still be live for 2 releases it must be tested
- add tests just to verify the new API works (a wild copy paste or a file duplication would be ok IMO)
- the old API should probably just call the new API:
```cpp
auto new_api(){ /* old implementation */ };
auto old_api(T unused_param){ return new_api();}
```

Did i answer your question ? 😄

andreacassioli:
yes. I hope the last question: shall I change the function naming? or just rely on the type of the parameters to distinguish the various overload?

Becheler:
Good question, always feel free to ask more !

I don't think you should change the name of the function, as the number and type of parameters will should the correct overload.

That being said, I do dislike the half-hungarian name. I would rather have it called `shortest_path_ressource_constrained` so it presents the core concept first, the detail after, and is more easily scannable, auto-completed and discoverable along with all the other `shortest_path_x_y_z` variations. But this is I believe not in the scope of the PR and would affect all the call sites (documentation, tests, examples ...).

@jeremy-murphy what do you think ? Should we jump on the occasion to rename the function?

Issue #447, opened 2026-01-10
[FEATURE] Louvain/Leiden Algorithm for Community Detection


Becheler:
Hi Graphies,

## Current state

If I understand correctly there is currently no method for community detection in BGL (but see clustering methods). Community detection algorithms like Louvain and Leiden automatically infer hidden groupings in very large networks, identifying gene modules, customer segments, fraud rings, or system components without requiring labeled data.
Common methods include:
- K-core Decomposition
- Label Propagation
- the Louvain algorithm (popular, ~30000 citations, some statistical issues)
- the Leiden algorithm (extends Louvain, solves some issues, more complex, less standard ~ 6000 citations)

## Scope

I plan to submit a PR to implement the simplest case of Louvain modularity optimization for undirected graphs only:
- minimal BGL-style API
- regression + correctness tests to reproduce validation methods/results in Blondel et al 2008
- no metric generalization yet

The benefit of starting with Louvain is that it builds requried components for Leiden, that is also more complex.

<img width="968" height="618" alt="Image" src="https://github.com/user-attachments/assets/41374d6b-e66c-4e0f-bf6a-9cdb2824eb0b" />
Algorithm figure from Blondel at al. 2008

## Litterature and context

Related papers and ressource:
- [Newman and Girvan, 2004](https://arxiv.org/pdf/cond-mat/0308217): paper for modularity metrics, with per-community version too
- [Blondel et al, 2008](https://arxiv.org/pdf/0803.0476): Louvain foundational paper using modularity metrics
- [Campigotto et al, 2014](https://arxiv.org/pdf/1406.2518): Generalizing Louvain to multiple quality metrics + C++ implementation using Strategy pattern (see reference 4)
- [Traag et al, 2019](https://www.traag.net/wp/wp-content/papercite-data/pdf/traag_leiden_algo_2018.pdf) : Leiden paper + Louvain background
- [Louvain Method CS class video](https://www.youtube.com/watch?v=QfTxqAxJp0U)
- [Comparative evaluation of community detection methods ](https://hal.science/hal-01976587/document) : methods, performances, generated community sizes distributions

## Formalism

Modularity defined as:
```math
Q = \frac{1}{2m} \sum_{ij} \left[ A_{ij} - \gamma \frac{k_i k_j}{2m} \right] \delta(c_i, c_j)
```
Where:
 - $m$ is the sum of all edges weights
- $A$ is the adjacency matrix
- $\gamma$ is the resolution parameter
[...]

PR #453, opened 2026-02-02
Add Louvain community detection algorithm

Becheler:
Implement #447

- Multi-level modularity optimization following Blondel et al. (2008)
- Supports custom quality functions (other than modularity) with policy-based design (extensions to come to propose alternative quality functions to match gen-louvain)
- Incremental quality tracking
- Lazy rollback
- Vertex shuffling
- Competitive with established implementation, see benchmarks [here](https://github.com/Becheler/boost-graph-benchmarks)

## Review-driven changes implemented in this PR:

- [x] Keep revision history / metadata in git (remove embedded revision-history metadata)
- [x] Define Boost concepts for user-provided graph partition quality functions (including incremental variant)
- [x] Allow users to provide a quality function (support both naive and incremental implementations)
- [x] Make RNG configurable by the user (URBG-style), with a default
- [x] Use BOOST_ASSERT instead of assert
- [x] Fix size()/loop index type mismatch (std::size_t)
- [x] Return vertex partitions as contiguous integral labels (match BGL interfaces)
- [x] Use index-based storage instead of hash maps to access vertices (generic across graph types)
- [x] Update the documentation to reflect those changes
- [x] Provide the documentation requested by Joaquin (louvain_clustering, louvain_quality_functions)
- [x] Rework unfold helper/type based on reviewer suggestion (free function in detail namespace; avoid _t suffix for classes)
- [x] Assert/check add_edge insertion success when using add_edge return value
- [x] After benchmarking for correctness and performance, I removed a Q failsafe (aggregation is basically modularity-preserving) and removed a best-partition tracking approach. Code is simpler and faster and as much correct.

## Review items still open:

- [ ] Make internal containers fully user-configurable via templating the map type (avoid hard-coding std::map/std::set; consider flat_map default)

I reviewed allocating container in the file (4 maps/sets and ~13 vectors). They all live in `louvain_detail` internals or are local scratch inside the public function, they do not appear in the API signature.

The maps  operate on types created by the internal `aggregate()` (`agg_vertex_t`, `community_type`) that the user never sees. The set only allocates in the slow/non-incremental fallback path. The final remap is O(k) and one-shot and exists only to give meaningful contiguous labels. The vectors back `iterator_property_map`, which requires random-access iterators.

All maps/sets already use `boost::unordered_flat_map`/`unordered_flat_set` so the `flat_map` suggestion is effectively already done.

Templating the container type would add a template parameter that leaks internal types into the public API for no user-observable benefit : the containers are transient scratch rebuilt each pass, not long-lived state.

In short I would rather keep the API simple for now and avoid making too much noise (the genericity level looks already fairly complex to me), but I would be happy to revisit if a concrete use case comes up. I would rather focus on a next PR about more direct benefits:

- [ ] Supporting undirected CSR graphs
- [ ] Supporting directed graphs (requires quality function tweaks)
- [ ] Supporting generic termination criterion as different application domains may require different stopping conditions, including fixed gain thresholds (Campigotto et al., 2014), threshold scaling (Halappanavar et al., 2017), decisions learned from gain decay patterns, or number of vertices moved, or a mix of those.


jeremy-murphy:
Not sure why this error is occurring only for OSX 11.7 C++14. I assume this is not specific to your code.
```
In file included from ../../../boost/container/detail/operator_new_helpers.hpp:26:
../../../boost/container/detail/aligned_allocation.hpp:87:26: error: no member named 'aligned_alloc' in the global namespace; did you mean 'aligned_allocate'?
   return rounded_size ? ::aligned_alloc(al, rounded_size) : 0;
                         ^~~~~~~~~~~~~~~
                         aligned_allocate
../../../boost/container/detail/aligned_allocation.hpp:77:14: note: 'aligned_allocate' declared here
inline void* aligned_allocate(std::size_t al, std::size_t sz)
             ^
1 error generated.
```

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,474 @@
...
+// Authors: Arnaud Becheler
+//
+// Distributed under the Boost Software License, Version 1.0. (See
+// accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+//=======================================================================
+//
+//
+// Revision History:
+//
```
jeremy-murphy: I know embedded revision histories were a thing in the past, but I would prefer to keep all this metadata in git.

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,474 @@
...
+{
+
+/// @brief Result of graph aggregation operation.
+template <typename Graph, typename PartitionMap, typename VertexDescriptor, typename WeightType>
+struct aggregation_result
+{
+    Graph graph;
+    PartitionMap partition;
+    std::map<VertexDescriptor, WeightType> internal_weights;
+    std::map<VertexDescriptor, std::set<VertexDescriptor>> vertex_mapping;
```
jeremy-murphy: We generally try to avoid hard-coding the choice of data structure, especially `std::map` and `std::set`, so instead of templating `VertexDescriptor` and `WeightType` we should template the whole map type, so users can use `boost::unordered_map` or some other kind of property map of their own choice.

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,474 @@
...
+    using edge_property_t = property<edge_weight_t, weight_type>;
+    using aggregated_graph_t = adjacency_list<vecS, vecS, undirectedS, no_property, edge_property_t>;
+    using result_t = aggregation_result<aggregated_graph_t, vector_property_map<vertex_descriptor>, vertex_descriptor, weight_type>;
+    
+    aggregated_graph_t new_g;
+    vector_property_map<vertex_descriptor> new_community_map;
+    
+    std::set<community_type> unique_communities;
+    std::map<community_type, vertex_descriptor> comm_to_vertex;
+    std::map<vertex_descriptor, std::set<vertex_descriptor>> vertex_to_originals;
```
jeremy-murphy: These should almost certainly be input parameters taken by non-const reference so that the user a) decides their type and b) automatically gets their value at the end.

jeremy-murphy (commented):
Just a few comments, more later.

include/boost/graph/louvain_clustering.hpp
Becheler: I did not like it either, I'm glad you don't ahah :)

include/boost/graph/louvain_clustering.hpp
Becheler: Mhhh I see. I felt it was not in the BGL spirit to do so, and Joaquin also mentioned it, but I was not sure how to solve it without making the API heavy. Can we default to a concrete type to simplify user's experience ? Also, is it ok to use `boost::unordered_map` if it adds the constraint of `key_type` being hashable ? That was my idea behind using `std::map`

include/boost/graph/louvain_clustering.hpp
Becheler: So they gain access to the whole hierarchy. Ufff it's a lot of guts leaking out haha 
Will do, and tell you in case of problemsm thanks again for your time !

jeremy-murphy: I could be wrong. But have a look at the `astar` API for some examples of prior art.

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,474 @@
+//=======================================================================
+// Copyright 2026 Becheler Code Labs for C++ Alliance
+// Authors: Arnaud Becheler
+//
```
joaquintides, line 3: You retain the copyright, so no need to mention the C++ Alliance. Also, is "Becheler Code Labs" a real legal entity? If this is not the case, then assigning (C) to your physical person would be best.

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,474 @@
...
+    
+    const level_t& operator[](std::size_t i) const {
+        return levels[i];
+    }
+    
+    template <typename CommunityMap>
+    auto unfold(const CommunityMap& final_partition) const
+    {
+        assert(!empty());
+
```
joaquintides, line 145: Not sure what the BGL convention is, Boost libs generally use `BOOST_ASSERT` instead.

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,474 @@
...
+        std::map<VertexDescriptor, VertexDescriptor> original_partition;
+        
+        for (const auto& kv : final_partition) {
+            std::set<VertexDescriptor> current_nodes;
+            current_nodes.insert(kv.first);
+            
+            // From coarse to fine
+            for (int level = size() - 1; level >= 0; --level) {
```
joaquintides: Type mismatch, `size()` is a `std::size_t`. More idiomatic, but maybe not to your taste:

```cpp
for(auto level = size(); level--; ) {
```

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,474 @@
...
+}
+
+}} // namespace boost::louvain_detail
+
+namespace boost 
+{
+
+template <typename Graph, typename CommunityMap, typename WeightMap, typename QualityFunction = newman_and_girvan>
+typename property_traits<WeightMap>::value_type
+louvain_local_optimization(
```
joaquintides: Is this function supposed to be public?

include/boost/graph/louvain_quality_functions.hpp
```diff
@@ -0,0 +1,304 @@
...
+    struct community_storage_selector<CommunityType, ValueType, false> {
+        using type = std::map<CommunityType, ValueType>;
+    };
+}
+
+// Modularity: Q = sum_c [ (L_c/m) - (k_c/2m)^2 ]
+// L_c = internal edge weight for community c
+// k_c = sum of degrees in community c
+// m = total edge weight / 2
+struct newman_and_girvan
```
joaquintides, line 177: If users can provide their own quality function other than `newman_and_girvan`, you probably need to define a concept `LouvainQuality` of sorts. Take a look at how's done at

https://github.com/boostorg/graph/blob/develop/include/boost/graph/astar_search.hpp#L56

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,474 @@
...
+    
+    // populates k[c], in[c], tot[c]
+    weight_type Q = QualityFunction::quality(g, communities, w, k, in, tot, m);
+    weight_type Q_new = Q;
+    std::size_t num_moves = 0;
+    std::size_t pass_number = 0;
+    bool has_rolled_back = false;
+    
+    // Randomize vertex order once
+    std::mt19937 gen(seed);
```
joaquintides: This sould be made generic and passed by the user (with a default arg) as a [URGB&&](https://en.cppreference.com/w/cpp/named_req/UniformRandomBitGenerator.html).

include/boost/graph/louvain_quality_functions.hpp
joaquintides, line 177: Is this modularity thing a general concept or does it apply to Louvain only?

include/boost/graph/louvain_clustering.hpp
jeremy-murphy: Oh yes, we can still (and should) make the user experience nice with defaults. That's the great thing, we get both benefits, the cost is more work on the part of the library authors. :)
Again, I think `astar` is an example, but instead of using `param =  arg` in the function definition, make an overload, like so: 

```
auto user_friendly_foo(graph const &g) {
ConcreteA a;
ConcreteB b;
return generic_foo(g, a, b);
}
```

Sorry, typing on my phone, so please ignore random syntax errors.

jeremy-murphy: Umm, yeah, we probably shouldn't add new constraints into the default interface. Users will too easily assume that the constraint is mandatory. 
So maybe use boost::flat_map as the default and users can always use a hash map if they want to.

include/boost/graph/louvain_clustering.hpp
jeremy-murphy: And on second thought, this is not a priority, we can always do it later. Getting it correct and fast are higher priorities.

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,474 @@
...
+        bool inserted;
+        tie(e, inserted) = add_edge(kv.first.first, kv.first.second, kv.second, new_g);
+    }
+    
+    return result_t{std::move(new_g), std::move(new_community_map), std::move(temp_internal_weights), std::move(vertex_to_originals)};
+}
+
+// Track hierarchy of aggregation levels for unfolding partitions.
+template <typename VertexDescriptor>
+struct hierarchy_t
```
jeremy-murphy: Only one data member and no real invariants to speak off suggests that you don't really need this type. 
I think `unfold` could be a free function, maybe in a private namespace depending on how reusable it is, that takes `levels`  and final_partition as parameters.

By the way, please only use the `_t` suffix for type aliases, not for names of classes.

include/boost/graph/louvain_clustering.hpp
jeremy-murphy: I don't really mind, but yeah, better to follow whatever is common practice.

jeremy-murphy (changes requested):
Couple more requests, still going...

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,474 @@
...
+            auto edge_key = std::make_pair(std::min(new_u, new_v), std::max(new_u, new_v));
+            temp_edge_weights[edge_key] += w;
+        }
+    }
+    
+    // add edges to connect super nodes
+    for (const auto& kv : temp_edge_weights) {
+        edge_descriptor e;
+        bool inserted;
+        tie(e, inserted) = add_edge(kv.first.first, kv.first.second, kv.second, new_g);
```
jeremy-murphy, line 127: If we're getting the result of `add_edge`, then presumably we should assert that the edge was inserted or something.

include/boost/graph/louvain_quality_functions.hpp
Becheler: This is a yes and no situation.
- The modularity can be used outside of louvain to assess partition quality of a graph.
- But the current implementation with incremental computations (`remove`, `insert`, `gain`) is particularly suited for Louvain.
- Making it generally useful would require to disentangle the two aspects. 
- But I would rather do it in a `clustering` folder so we can have:
```
include/boost/graph/clustering/
├── quality_functions.hpp        # 10 incremental metrics/criterions for gen-louvain
├── label_propagation.hpp      # another clustering method (future work)
├── leiden.hpp                          # Leiden algorithm (future work)
├── louvain.hpp                        # Louvain algorithm
├── girvan_newman.hpp         # Edge betweenness clustering (currently in bc_clustering.hpp)
```

include/boost/graph/louvain_clustering.hpp
Becheler: I was not too sure, as it can theoretically be used as a one pass optimization step. But you're right it's obviously not the main usage, so I move it to details :)

include/boost/graph/louvain_clustering.hpp
Becheler: Done 👍🏽

include/boost/graph/louvain_clustering.hpp
Becheler: Done 👍🏽

include/boost/graph/louvain_clustering.hpp
Becheler: Done 👍🏽

include/boost/graph/louvain_clustering.hpp
Becheler: Done 👍🏽

include/boost/graph/louvain_clustering.hpp
Becheler: Done 👍🏽

include/boost/graph/louvain_clustering.hpp
Becheler: Done 👍🏽

include/boost/graph/louvain_quality_functions.hpp
Becheler: Done 👍🏽 
I defined:
- `GraphPartitionQualityFunctionConcept`
- `GraphPartitionQualityFunctionIncrementalConcept`

include/boost/graph/louvain_clustering.hpp
Becheler: Done 👍🏽

doc/louvain_clustering.html
```diff
@@ -0,0 +1,248 @@
...
+
+IN: <tt>const WeightMap&amp; w</tt>
+<blockquote>
+  Edge weights. Must model
+  <a href="../../property_map/doc/ReadablePropertyMap.html">Readable
+  Property Map</a> with the graph's edge descriptor as key type.
+  Weights must be non-negative.
+</blockquote>
+
+IN: <tt>URBG&amp;&amp; gen</tt>
```
joaquintides, line 121: Would it make sense to provide a default arg for this?

doc/louvain_clustering.html
```diff
@@ -0,0 +1,248 @@
...
+<blockquote>
+  A random number generator used to shuffle the vertex processing
+  order at each pass. Any type meeting the C++
+  <i>UniformRandomBitGenerator</i> requirements works
+  (e.g.&nbsp;<tt>std::mt19937</tt>).
+</blockquote>
+
+IN: <tt>weight_type min_improvement_inner</tt>
```
joaquintides, line 136: Is Louvain guaranteed to converge (i.e. to eventually stop)? If this is not clear, does it make sense to provide additional hard limits on the number of inner/outer iterations?

doc/louvain_clustering.html
```diff
@@ -0,0 +1,248 @@
...
+    compile time and selects the appropriate code path automatically.
+
+<H3>Where Defined</H3>
+
+<P>
+<a href="../../../boost/graph/louvain_clustering.hpp"><TT>boost/graph/louvain_clustering.hpp</TT></a>
+
+<H3>Parameters</H3>
+
+IN: <tt>const Graph&amp; g</tt>
```
joaquintides, line 90: The algorithm has the additional requirement that vertices are copyable, hashable etc., as they're internally stored in unordered_sets.

doc/louvain_clustering.html
Becheler: I am not sure what it would be. And also it would differ from what i've seen in `random.hpp` utilities:
https://github.com/boostorg/graph/blob/3131c24630e42c79b43c1f32558041c219ab84b8/include/boost/graph/random.hpp

doc/louvain_clustering.html
Becheler: If I understand well it's guaranteed to terminate in theory:
- the quality (modularity) is monotically non-decreasing through the algorithm
- becasue the algorithm is supposed to only accept nodes moves and community merges that strictly improve (or does not decrease in some variations) the quality of the partition
- modularity is bounded <=1
- the number of partition is finite

That being said there is the case of large graphs and the trouble on floating point precision.

- For very large graphs maybe it would make sense to have some sense of async task: "please dear louvain, aggregate this for some time and when I'm done with waiting give me the last aggreagated graph you had". But that sounds like a very different interface ?
- The current API is still more flexible than igraph and genlouvain, that do not offer parametrization of the stopping condition (igraph has 0 for inner and outer thresholds and genlouvain has 10-6, fixed)

Am i making sense ?

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,630 @@
...
+#include <boost/assert.hpp>
+#include <boost/static_assert.hpp>
+#include <boost/type_traits/is_same.hpp>
+#include <boost/concept/assert.hpp>
+
+#include <boost/unordered/unordered_flat_map.hpp>
+#include <boost/unordered/unordered_flat_set.hpp>
+#include <boost/container_hash/hash.hpp>
+#include <algorithm>
+#include <iostream>
```
joaquintides: Is this `#include` needed?

doc/louvain_clustering.html
Becheler: You're right. I have been changing the vertices handling in this aspect because it was not friendly with some types of graphs. The interface now takes a VertexIndexMap but I still have to commit those changes, sorry 😓 
I will update the documentation in that sense once I merged the new stuff

include/boost/graph/louvain_clustering.hpp
Becheler: Hem. Nope 😄

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,630 @@
...
+
+// Hash specialization for std::pair to use with boost::unordered containers
+namespace std {
```
joaquintides: Big no-no :-)
* It's forbidden to specialize hash for types other than user-defined ones, which `std::pair` is not.
* Boost.Unordered does not use `std::hash` by default, but `boost::hash`. Your `temp_edge_weights` works because `boost::hash` works off the shelf with pairs, so the `std` specialization is not even used.

Becheler: Oooooh 🥲 Thank you, will remove

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,630 @@
...
+                has_rolled_back = true;
+            }
+        }
+        
+    } while (num_moves > 0 && (Q_new - Q) > min_improvement_inner);
+    
+    return Q_new;
+}
+
+/// @brief Fast version, requires the QualityFunction to implement GraphPartitionQualityFunctionIncrementalConcept
```
joaquintides: Looks like this comment is wrong, this version does _not_ require the quality function to be incremental.

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,630 @@
...
+){
+    using is_incremental = louvain_detail::is_incremental_quality_function<QualityFunction, Graph, CommunityMap, WeightMap>;
+    return local_optimization_impl<QualityFunction>(g, communities, w, std::forward<URBG>(gen), min_improvement_inner, is_incremental{});
+}
+
+} // namespace louvain_detail
+
+/// @brief Find the best partition of the vertices of a graph conditionally to a quality function like modularity.
+/// @return the modularity value of the best partition.
+template <typename QualityFunction = newman_and_girvan, typename Graph, typename ComponentMap, typename WeightMap, typename URBG>
```
jeremy-murphy, line 506: I'll have to dig into why, but the convention in C++ is to pass callbacks as objects, not types, and use `operator()`. I.e.:

```
template <typename Graph, typename BinaryFunction>
void foo(Graph g, BinaryFunction f)
{
  // ...
  f(g);
  // ...
}
```

And then, as I think the rest of BGL follows, you would choose a default value with
```
template <typename Graph>
foo(Graph g)
{
  return foo(g, your_default);
}

```

jeremy-murphy: I'm not sure there is fundamental reason why, but basically because `std::function` is the archetypal callback that can wrap anything, that's why "an object with `operator()`" is the convention.

jeremy-murphy (changes requested):

doc/louvain_clustering.html
jeremy-murphy: ### Incremental output

There is a general interface for keeping only the last value, but it's not a pattern I've actually ever seen in practice.

Instead of returning a single value at the end (either by return value or in/out parameter), emit values by output iterator as they are calculated. Now, normally values emitted by output iterator are all kept by pushing back on a vector for example. But with a "clobbering" output iterator that always writes to the same location, the user can choose to just keep the last value. Neat, huh?

However, that kind of updating output value is only necessary if we do actually allow the user to specify an early termination condition. (Which, by the way, is orthogonal to asynchrony. At least for now, until we start supporting C++20 and start returning futures from algorithms. If the user wants to run the algorithm in the background, they use `std::async`.) Usually that kind of user-specified condition should come in the form of a callback predicate, so the user can either use number of iterations or time elapsed and it is opaque to the algorithm. The algorithm would provide a default predicate that does whatever is the most reasonable thing to do, which might be applying no early termination condition at all.
### Floating-point
This is a huge topic, but in short we don't want to let floating-point accuracy cause our algorithm to run indefinitely. The basics are a) using a robust summation algorithm such as Kahan's, which is in Boost.Accumulators, and ... I can't remember what else.  :) Looking through the calculations to find subtractions that could be between two extremely close values. I might be able to provide more ideas later.

include/boost/graph/louvain_clustering.hpp
Becheler: Yes I also wondered about making the qualities functors. I guess because copiable + callable is the kind of minimal constraint one can think about. Just to be clear, are you asking about change or are you still evaluating it ? :)

jeremy-murphy: Oh yes, you should change it to match the convention.

Becheler: Although I agree, one setback is that there is one more argument to pass to the function, making defaulting values tricky as more and more are added :)

include/boost/graph/louvain_clustering.hpp
Becheler: fixed :)

include/boost/graph/louvain_clustering.hpp
Becheler: done :)

include/boost/graph/louvain_clustering.hpp
Becheler: done :)

include/boost/graph/louvain_clustering.hpp
jeremy-murphy: We'll find a way. ;)

jeremy-murphy:
Btw, are the benchmarks in the description current? I mean, did the performance change at all since you swapped some standard map containers for Boost ones?
I'm curious to know if it made a difference but in the description we just want whatever is the latest benchmark.

doc/louvain_clustering.html
Becheler: Thank you for your ideas ! 😄 
That sounds like matter for a next iteration on the algorithm, right ? Right now the termination criteria is fairly simple and avoids the issue of floating point precision ( Qlast - Qnow < threshold) so we should be ok. 
The only thing it affects is that on a same graph, brute force quality computation and incremental computations lead to somewhat different values:
image

It's still not clear how this difference in partition quality computation impacts the underlying partition optimization path, but I doubt it is very impacting (algthough I should check this out, but then I would need to find a metric to compare partitions, like Normilized Mutualize Information or Variation of Information

Becheler:
@jeremy-murphy 
- I have this benchmark repository here: https://github.com/Becheler/boost-graph-benchmarks/tree/main/louvain
- they are run locally then results pushed, but we chatted with Joaquin about making it a CI thing
- switching hashmaps did not improve performance, but I have been switching to indexing anyway because it was more generic for some types of graphs
- I haven't done a full benchmark since the last changes (its a few hours of computation lol) so the figures are somewhat outdated 
- The seemingly really fast performance gain between genlouvain and igraph comes from a different tradeoff between termination criterion and partition quality:
    - I compared bgl 0 threshold against genlouvain 10-6 threshold.
    - I aligned their behavior along the same total number of passes
    - the decreasing curves answer _"what percentage of total gain is still unrealized after n passes"_
    - the little bars answer _"how much this pass contributed"_
    - and you can see the 5 levels of aggregation as the algorithm tries to find the best partition into communities
    - what is interesting is that local optimization in the first level eats most of the passes for very little gain.
    - `min_inner_threshold` affect the number of passes spent inside each level
    - `min_outer_threshold` affect the total number of levels explored
    - And so the graph shows that (for this graph case) `min_inner_threshold` is the main driver of the number of iterations:
image

include/boost/graph/louvain_clustering.hpp
Becheler: So I began to implement it, did not feel it and made a compromise, tell me if it's ok with you.

A quality function is not a single-action callable: it is a policy object with five distinct operations: full modularity computation, modularity from pre-computed aggregates, incremental gain, and aggregate updates for vertex removal and insertion. 

Elevating any one to `operator()` felt arbitrary, and overloading all five is not viable since some share the same arity with different semantics.

Instead i tried to adapt the established BGL visitor pattern instead: named member functions (`qf.quality(...)`, `qf.gain(...)`, …) called on a `QualityFunction` object parameter, as `BFSVisitor` exposes `discover_vertex`, `examine_edge`, etc.

I passed the policy as a copy because they are supposed to be cheap to copy. Does it work for you ?

jeremy-murphy:
Btw, you can use - [ ]  and - [x] in the markup to make a todo list and tick things off as they're done. It will make it easier for us as reviewers to know where you are up to. E.g.:

- [x] done
- [ ] not done

include/boost/graph/louvain_clustering.hpp
jeremy-murphy: Sounds right in principle, let me look at the code...

jeremy-murphy: There is something odd about having a default value for the template parameter. Maybe it's fine, but maybe just writing `QualityFunction f = newman_and_girvan{}` is better. It's fine either way, so I'll resolve this thread now.

doc/louvain_clustering.html
jeremy-murphy: It's not a big deal and I'm happy to do it later, but users do complain about too many parameters in general, so where we can provide defaults, it is helpful. I.e., not providing default values is probably a tradition that needs breaking.

doc/louvain_clustering.html
jeremy-murphy: I think the algorithm sufficiently guarantees convergence. Floating-point accuracy is notoriously difficult to get right, so I'm happy to improve it over time.

doc/louvain_clustering.html
jeremy-murphy: Is this still an open issue or resolved?

include/boost/graph/louvain_clustering.hpp
```diff
@@ -0,0 +1,648 @@
...
+#ifndef BOOST_GRAPH_LOUVAIN_CLUSTERING_HPP
+#define BOOST_GRAPH_LOUVAIN_CLUSTERING_HPP
+
+#ifndef BOOST_GRAPH_LOUVAIN_TRUST_AGGREGATED_Q
+#define BOOST_GRAPH_LOUVAIN_TRUST_AGGREGATED_Q 0
+#endif
+
+#ifndef BOOST_GRAPH_LOUVAIN_TRACK_PEAK_Q
+#define BOOST_GRAPH_LOUVAIN_TRACK_PEAK_Q 0
+#endif
```
jeremy-murphy: Ahh, I only just discovered these macros, sorry, and I'm just wondering if they are just meant for private experimental purposes or if users would want to use them too?

If they're for private experiments, I think I would prefer they were kept on a private branch. I mean, I think we should only merge into `develop` the code that is useful to everyone.

If they _are_ useful to everyone, then they need to be documented and I don't think macros are really OK.

I think once this is resolved, then we're good to merge.  :)

Becheler: Hi @jeremy-murphy :)
Well you were pretty fast to notice because I just created them yesterday. The idea is that they help me benchmarking alternative designs to simplify the remaining code. The usage of macros was recommended by Joaquin because they are meant to disappear once I made a choice. My wrong was surely pushing them on the PR (but I use this branch for the benchmarks too and I did not want to mess up several repos).
Whatever I still have two little macros to test today and then I should be good and I will remove them (and some code paths along!)
Wahooo can't wait for the meeerge <3

jeremy-murphy:
This is essentially good to go, but I want to merge the what's currently on `develop` into `main` as soon as the Boost release window opens, then merge this into `develop` so that it's got some time to marinate.

include/boost/graph/louvain_quality_functions.hpp
```diff
@@ -0,0 +1,402 @@
...
+    /// @brief Hashable types use unordered_flat_map: get/put O(1) average, better cache locality
+    template <typename CommunityType, typename ValueType>
+    struct community_storage_selector<CommunityType, ValueType, true> {
+        using type = boost::unordered_flat_map<CommunityType, ValueType>;
+    };
+    
+    /// @brief Non-hashable types use flat_map: get/put O(1) average
+    template <typename CommunityType, typename ValueType>
+    struct community_storage_selector<CommunityType, ValueType, false> {
+        using type = boost::unordered_flat_map<CommunityType, ValueType>;
```
jeremy-murphy, line 68: This looks wrong... presumably a non-hashable `CommunityType` should use an ordered map like `std::map`. I'm not sure what an example of a non-hashable `CommunityType` is off the top of my head and I presume they're uncommon, so it's not super important to get non-hashable types working. 
We don't need to fix this before merging to `develop` as the hashable types work fine, but we should work resolve this before merging the whole feature to `main`.

Becheler: Would something like this answer the problem?

```cpp
template <typename T, typename = void>
struct is_hashable : std::false_type {};

template <typename T>
struct is_hashable<T, boost::void_t<
    decltype(std::declval<boost::hash<T>>()(std::declval<T const&>()))
>> : std::true_type {};

template <typename CommunityType, typename ValueType>
struct community_storage_selector<CommunityType, ValueType, false> {
    using type = std::map<CommunityType, ValueType>;
```

Note that I also struggle finding an example of a non-hashable community label. Boost::hash support pairs and tuples. Should we simply drop the trait ?

joaquintides: FWIW, this trait is wrong: `boost::hash<T>` is defined for any `T`, and only fails at `operator()` instantiation time if there's no `hash_value` overload for `T`, so your `is_hashable` always resolves to `true_type`.

Becheler: Thank you @joaquintides ! 
Well looks like I would better delete this useless and broken piece of code :)

PR #460, opened 2026-03-08
new figure aligned with the docs

### LINKED ISSUE #185: [boykov-kolmogorov] mismatch between doc/figure on colors
In the lib documentation,
https://www.boost.org/doc/libs/1_70_0/libs/graph/doc/boykov_kolmogorov_max_flow.html
the figure shows a white  Source node and a black Sink node, whether it is previously written that:
> the source and the sink are colored (source==black, sink==white)

This is confusing. Either change the picture or the code/doc.

IMO the source should be white and not black (i.e change the code and doc), but it would be too complicated to keep backward compatibility...

andreacassioli:
Fixes https://github.com/boostorg/graph/issues/185

I made a new figure identical to the old except that\

- source and sink are swapped in color as per the docs
- the image is now a png as most of the rest of the docs


jeremy-murphy:
Thank you! I see what you mean about the black/white inversion.

How did you make the diagram? I'm curious how you got that hand-drawn style and whether changing it would be easy or difficult.  😅 I personally would prefer to stick with a formal style, and preferably generated by Graphviz so that they are easy to reproduce, although that is a secondary consideration.

andreacassioli:
I have used this tool https://excalidraw.com/. Indeed it is not standard, but it takes 2 minutes!

I did not find any guideline and I could not see graphviz files for the other images, so I assumed people just made figures as needed.

jeremy-murphy:
I make a lot of guidelines up as I go. 😆 One day I will put them in the README.

I'll merge this since it does fix a glaring logical discrepancy and we can worry about style some other time.

jeremy-murphy:
Thanks!

PR #469, opened 2026-04-11
Fix circle layout docs and cleanup

andreacassioli:
Fix synopsis in the circle_layout docs ([currently broken](https://www.boost.org/doc/libs/latest/libs/graph/doc/circle_layout.html))

Incidentally I looked at the source code of that function and decided to do some small tweaking of the source code to add a const, use auto and remove a typedef.


andreacassioli:
any feedback on this tiny fix @jeremy-murphy

include/boost/graph/circle_layout.hpp
```diff
@@ -42,13 +41,10 @@ void circle_graph_layout(
...
 
-    typedef typename graph_traits< VertexListGraph >::vertices_size_type
-        vertices_size_type;
+    const auto n = num_vertices(g);
 
-    vertices_size_type n = num_vertices(g);
-
-    vertices_size_type i = 0;
-    double two_pi_over_n = 2. * pi / n;
+    auto i = 0;
```
Becheler: I believe `auto i = 0` is a slight regression on type-correctness. It deduces `int`, where the original code used the graph's `vertices_size_type` On graphs with more than INT_MAX vertices this would overflow ?

With `vertices_size_type`, `++i` cannot overflow because `i` only reaches `n - 1`, and `n` came from `num_vertices(g)` that returns `vertices_size_type`. By construction, `i` stays within the representable range of its type.

Becheler (changes requested):
I approved after the regression was fixed, I think it can be safely merged.
Let's wait for @jeremy-murphy 😄 
Plan to merge in order: #469 → #489  → #491

include/boost/graph/circle_layout.hpp
andreacassioli: reverted to the `vertices_size_type`

Becheler (approved):

jeremy-murphy (approved):

jeremy-murphy:
Sorry, thought I merged this already but I guess I only hit approve. 😅

PR #488, opened 2026-05-05
Enable adding Ubuntu toolchain PPA by default

Flamefire:
The default in Boost.CI will be false but some jobs still rely on this, so keep it for now the default.


jeremy-murphy:
Thanks!

Issue #496, opened 2026-06-05
State of warnings in CI `develop`


Becheler:
# Compiler warnings (global, full CI matrix)

Source: CI run 26759013540 on `develop` (push, sha `6efc6b60`, 2026-06-01). All 24 jobs succeeded; the counts below are warnings, not errors.

## Per-job warning counts

| Job | Warnings |
|-----|----------|
| windows_msvc_14_3 (msvc-14.3) | 1226 |
| macos (clang, 14) | 88 |
| macos (clang, 17) | 83 |
| macos (clang, 20) | 83 |
| ubuntu (clang-19, 14/17/20/23) | 76 each |
| ubuntu (gcc-14, 14/17/20/23) | 52 each |
| all `*-cmake-install` / `*-cmake-subdir` jobs | 0 |

The cmake jobs build only the library, so they surface nothing. All warnings come from the b2 toolset jobs that also build tests and examples. Counts are stable across language-standard variants (14/17/20/23), so the standard level does not change the warning set.

## Correctness-relevant (fix or investigate, do not just silence)

| File:line | Warning | Compilers |
|-----------|---------|-----------|
| detail/adjacency_list.hpp:2276 | variable `p` used uninitialized / passed as uninitialized const reference | clang (`-Wuninitialized-const-reference`) **and** gcc (`-Wmaybe-uninitialized`) |
| test/r_c_shortest_paths_test.cpp:160, :264 | possibly dangling reference to a temporary | gcc (`-Wdangling-reference`) |
| example/r_c_shortest_paths_example.cpp:158 | possibly dangling reference to a temporary | gcc (`-Wdangling-reference`) |
| pending/detail/disjoint_sets.hpp:58, :59 | signed/unsigned comparison | gcc (`-Wsign-compare`) |

`adjacency_list.hpp:2276` is the strongest signal: two independent compilers agree `p` may be read before initialization.

## Cosmetic clang/gcc warnings (dead code / deprecation)

Library headers:

| File:line | Warning | Flag |
|-----------|---------|------|
| howard_cycle_ratio.hpp:360 | lambda capture `this` not used | `-Wunused-lambda-capture` |
| r_c_shortest_paths.hpp:200 | unused typedef `LTraits` | `-Wunused-local-typedef(s)` |
| louvain_clustering.hpp:58, :226, :529 | unused type aliases | `-Wunused-local-typedef(s)` |
[...]

PR #499, opened 2026-06-06
remove narrowing conversion in `king_ordering.hpp`

Becheler:
Refs #496 

### Before submitting

### Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

Two commits:

1. Harden `king_ordering.cpp` unit test (it did not assert anything) to prevent regressions
2. Convert `king_ordering.hpp` heap from `int` to `std::ptrdiff_t` so the `size_t` queue positions no longer narrow

### Motivation

Refs #496 

### Testing

Same CI, more hardened tests, less warnings:

| Job | Baseline (develop) | After fix | Δ |
|-----|--------------------|---------------|---|
| MSVC 14.3 | 1226 | 1154 | −72 |
| ubuntu gcc-14 (14/17/20/23) | 51 | 51 | 0 |
| ubuntu clang-19 (14/17/20/23) | 75 | 75 | 0 |
| macOS clang 14 | 75 | 75 | 0 |
| macOS clang 17 | 70 | 70 | 0 |
| macOS clang 20 | 70 | 70 | 0 |
| all cmake jobs | 0 | 0 | 0 |

### Checklist

### DIFF
--- a/include/boost/graph/king_ordering.hpp
+++ b/include/boost/graph/king_ordering.hpp
@@ -14,6 +14,7 @@
 #include <deque>
 #include <vector>
 #include <algorithm>
+#include <cstddef>
 #include <boost/config.hpp>
 #include <boost/tuple/tuple.hpp>
 #include <boost/graph/detail/sparse_ordering.hpp>
@@ -33,7 +34,7 @@ namespace detail
     {
     public:
         bfs_king_visitor(OutputIterator* iter, Buffer* b, Compare compare,
-            PseudoDegreeMap deg, std::vector< int > loc, VecMap color,
+            PseudoDegreeMap deg, std::vector< std::ptrdiff_t > loc, VecMap color,
             VertexIndexMap vertices)
         : permutation(iter)
         , Qptr(b)
@@ -60,9 +61,10 @@ namespace detail
             // heap the vertices already there
             std::make_heap(rbegin, rend, [this](const auto& first, const auto& second) { return comp(second, first); });
 
-            unsigned i = 0;
+            std::ptrdiff_t i = 0;
 
-            for (i = index_begin; i != Qptr->size(); ++i)
+            for (i = index_begin; i != static_cast< std::ptrdiff_t >(Qptr->size());
+                 ++i)
             {
                 colors[get(vertex_map, (*Qptr)[i])] = 1;
                 Qlocation[get(vertex_map, (*Qptr)[i])] = i;
@@ -101,10 +103,10 @@ namespace detail
 
     protected:
         // this function replaces pop_heap, and tracks state information
-        template < typename Vertex > void percolate_down(int offset)
+        template < typename Vertex > void percolate_down(std::ptrdiff_t offset)
         {
-            int heap_last = index_begin + offset;
-            int heap_first = Qptr->size() - 1;
+            std::ptrdiff_t heap_last = index_begin + offset;
+            std::ptrdiff_t heap_first = Qptr->size() - 1;
 
             // pop_heap functionality:
             // swap first, last
@@ -114,21 +116,21 @@ namespace detail
             std::swap(Qlocation[heap_first], Qlocation[heap_last]);
 
             // set drifter, children
-            int drifter = heap_first;
-            int drifter_heap = Qptr->size() - drifter;
+            std::ptrdiff_t drifter = heap_first;
+            std::ptrdiff_t drifter_heap = Qptr->size() - drifter;
 
-            int right_child_heap = drifter_heap * 2 + 1;
-            int right_child = Qptr->size() - right_child_heap;
+            std::ptrdiff_t right_child_heap = drifter_heap * 2 + 1;
+            std::ptrdiff_t right_child = Qptr->size() - right_child_heap;
 
-            int left_child_heap = drifter_heap * 2;
-            int left_child = Qptr->size() - left_child_heap;
+            std::ptrdiff_t left_child_heap = drifter_heap * 2;
+            std::ptrdiff_t left_child = Qptr->size() - left_child_heap;
 
             // check that we are staying in the heap
             bool valid = (right_child < heap_last) ? false : true;
 
             // pick smallest child of drifter, and keep in mind there might only
             // be left child
-            int smallest_child = (valid
+            std::ptrdiff_t smallest_child = (valid
                                      && get(degree, (*Qptr)[left_child])
                                          > get(degree, (*Qptr)[right_child]))
                 ? right_child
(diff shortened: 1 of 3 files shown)


include/boost/graph/king_ordering.hpp
```diff
@@ -189,19 +194,19 @@ namespace detail
 
                 child_location = parent_location;
                 heap_child_location = heap_parent_location;
-                heap_parent_location = (int)(heap_child_location / 2);
+                heap_parent_location = (std::ptrdiff_t)(heap_child_location / 2);
```
jeremy-murphy, line 197: This cast is redundant.
```suggestion
                heap_parent_location = heap_child_location / 2;
```

test/king_ordering.cpp
```diff
@@ -107,11 +128,11 @@ int main(int, char*[])
...
         for (size_type c = 0; c != inv_perm.size(); ++c)
             perm[index_map[inv_perm[c]]] = c;
-        std::cout << "  bandwidth: "
-                  << bandwidth(G,
-                         make_iterator_property_map(
-                             &perm[0], index_map, perm[0]))
-                  << std::endl;
+        size_type bw = bandwidth(
+            G, make_iterator_property_map(&perm[0], index_map, perm[0]));
+        std::cout << "  bandwidth: " << bw << std::endl;
```
jeremy-murphy, line 133: I would just remove all the console output as it's just noise 99% of the time. If we need console output only occasionally, then it should use a logger with configurable output levels so that a developer can easily turn it on during development.

So wait... was this test not actually testing anything, just printing console output? 😅

test/king_ordering.cpp
```diff
@@ -69,7 +70,27 @@ int main(int, char*[])
     property_map< Graph, vertex_index_t >::type index_map
         = get(vertex_index, G);
 
-    std::cout << "original bandwidth: " << bandwidth(G) << std::endl;
+    const size_type n = num_vertices(G);
+    auto assert_valid_permutation
+        = [&index_map, n](const std::vector< Vertex >& order) {
+              BOOST_TEST(order.size() == n);
+              std::vector< bool > seen(n, false);
```
jeremy-murphy, line 77: This particular use is OK, but in general prefer `boost::dynamic_bitset`.

include/boost/graph/king_ordering.hpp
```diff
@@ -164,17 +166,20 @@ namespace detail
...
         {
 
-            int child_location = Qlocation[vertex];
-            int heap_child_location = Qptr->size() - child_location;
-            int heap_parent_location = (int)(heap_child_location / 2);
-            unsigned parent_location = Qptr->size() - heap_parent_location;
+            std::ptrdiff_t child_location = Qlocation[vertex];
+            std::ptrdiff_t heap_child_location = Qptr->size() - child_location;
+            std::ptrdiff_t heap_parent_location
+                = (std::ptrdiff_t)(heap_child_location / 2);
```
jeremy-murphy, line 176: This cast is redundant.
```suggestion
            std::ptrdiff_t heap_parent_location = heap_child_location / 2;
```

include/boost/graph/king_ordering.hpp
```diff
@@ -33,7 +34,7 @@ namespace detail
     {
     public:
         bfs_king_visitor(OutputIterator* iter, Buffer* b, Compare compare,
-            PseudoDegreeMap deg, std::vector< int > loc, VecMap color,
+            PseudoDegreeMap deg, std::vector< std::ptrdiff_t > loc, VecMap color,
```
jeremy-murphy, line 37: Why do you think this is `std::ptrdiff_t`?
I've looked at the code, and I'm having trouble figuring out what it should be, but I'm inclined to think it should be `std::size_t`, unless there is a requirement for it to be signed?

jeremy-murphy (commented):
Definite improvement to have proper unit tests, but a few comments and questions about code choices.  :)

test/king_ordering.cpp
Becheler: Agreed, I was surprised at the unit test sending to console output. It was basically checking compilation + absence of hard crash.

- To be fair, I came to believe this ordering is a non trivial thing to test: I don't think we can guarantee exact expected ordering so if we do we may have flickering CI.
- I wanted to preserve whatever behavior we had from the medieval period of BGL, so I favored not touching anything but adding an invariant (the permutation validation): I can still copy paste the unit test to CE and observe its output and compare it to the expected outputs in the top comment (interestingly i had the two first IDs switched that seems to confirm we may not be able to freeze ordering expectations)
- removing the output would have removed the only way to guarantee some form of algorithmic expectation
- adding a logger would have brought much more noise than this PR was supposed to be about.

But happy to revisit unit tests outputs in a dedicated PR 😄

test/king_ordering.cpp
Becheler: yes I felt weird writing this vector of bool and tripled checked online it was fine here ahaha. Happy to learn about `boost::dynamic_bitset`!

include/boost/graph/king_ordering.hpp
Becheler: oops yup, the original cast was already redudant, this just got carried along. I will drop it.

include/boost/graph/king_ordering.hpp
Becheler: oops yup, the original cast was already redudant, this just got carried along. I will drop it.

include/boost/graph/king_ordering.hpp
Becheler: The signed requirement comes from the arithmetic the heap is doing:
https://github.com/boostorg/graph/blob/5c61bd0bb01bde368dc6c1aa99a0e449445d188a/include/boost/graph/king_ordering.hpp#L119-L123
and then:
https://github.com/boostorg/graph/blob/5c61bd0bb01bde368dc6c1aa99a0e449445d188a/include/boost/graph/king_ordering.hpp#L128-L129

For context, `king_ordering` repeatedly needs the lowest-pseudo-degree (= count of unnumbered neighbors, mutable through the algo) vertex among the current candidates. Hence the priority queue with a decrease-key: `percolate_down` is the `extract-min`, `percolate_up` is the `decrease-key` fix-up after a neighbor's pseudo-degree drops. The hand-rolled version also tracks each vertex's current slot in `Qlocation` so it can find and re-heapify a vertex when its pseudo-degree changes.

So at each step the algorithm takes the unnumbered vertex with the fewest remaining unnumbered neighbors. King_ordering lays its heap out backwards (root at the high index, growing toward 0), so for them off-the-end comes out negative instead of too-big.

That negative is why the index has to be signed. A child slot is computed as `Qptr->size() - right_child_heap`, and for a leaf that subtraction is meant to go negative. The guard right after (`valid = (right_child < heap_last)`) reads the negativity as "off the heap, stop". With `size_t` the same subtraction wraps to a huge value instead, the guard passes, and we index `(*Qptr)[right_child]` out of bounds.

Why `ptrdiff_t` and not the original `int`: `int` is signed but 32-bit on MSVC while `size()` is 64-bit, so the size-derived values narrow into it (that's the C4267/C4244 we're clearing). `ptrdiff_t` is just the signed type that's also `size()`'s width, so it keeps the sign and drops the warning.

I hope this makes sense. I've spent quite some time gathering context around king_ordering and the heap implementation, so im reasonably confident this is a bit more correct, but details are not very sharp in my mind so don't take this as an expert feedback either 😄

Issue #493, opened 2026-05-23
Implementing personalized PageRank for graph node scoring.


maniospas:
### Before filing

- [x] I searched the [existing issues](https://github.com/boostorg/graph/issues?q=is%3Aissue) and did not find a duplicate.
- [x] This is a concrete proposal, not an open-ended design question.

### Kind of addition

- [x] New algorithm
- [ ] New data structure
- [ ] New property map or utility
- [ ] New I/O format reader or writer
- [ ] Extension to an existing component
- [ ] Performance improvement
- [ ] Other (specify below)

### Motivation

Following preliminary discussion in #492 , this proposal aims to bring the popular personalized node scoring/ranking algorithm to BGL. Personalized node scoring can be used to score nodes with respect to their structural proximity to a (weighted) set of seed nodes, so it can be used for overlapping community member scoring and ranking. It can also be used as a building block for more complex community detection schemes that satisfy theoretical properties like finding structural communities of low conductance given that the seed nodes lie closely together. Finally, node scores can be used for link prediction by starting from one node and finding structural close alternatives.

The proposal below presents some basic information for bringing these kinds of methods, and can be used as future basis for also implementing more spectral graph filters (not to be confused with full-blown spectral processing in that spectral filters act on eigenvalue-based properties by actually working only on the vertex domain for near-linear instead of quadratic time complexity).

### Proposed addition

In linear algebra terms, the algorithm implements the iterative scheme:

$p = (1-dmp)Ap+dmp p_0$

where A is broadly a normalized graph adjacency matrix, dmp is the equivalent of the dampening factor (equal to 1-restart probability of equivalent random walk with restart schemes), and $p_0$ a starting personalization vector. As an API, this would mainly follow the existing `page_rank` interface:

```cpp
template <typename Graph, typename RankMap, typename Done>
void personalized_page_rank(const Graph& g, RankMap rank_map, Done done,
               typename property_traits<RankMap>::value_type damping,
               typename graph_traits<Graph>::vertices_size_type n,
               Normalization normalization,
               Optimizer optimizer,
               RankMap1 rank_map1,
               RankMap2 rank_map2);
```

Simplifications analogous to `page_rank`'s will be added, for example so that `done` uses some common default (see below for planned options for convergence tracking; it is not as simple a problem as it appears).

**Important differences:**

[...]

PR #491, opened 2026-05-19
Doc migration

Becheler:
## Migrate Boost.Graph documentation to Antora

NOTE: There is no way to split this PR properly, as all `.html` files need to be replaced all at once.
NOTE: The ~40000 lines of change in doc source code don't need direct review. Instead please review the preview currently hosted on the fork's Github Pages: https://becheler.github.io/graph/graph/index.html

This PR switches Boost.Graph from the pre-baked BoostBook/Quickbook HTML at `doc/*.html` to an Antora source tree under `doc/modules/ROOT/`. Once merged, the next Boost release will publish the new site at `www.boost.org/doc/libs/<version>/libs/graph/` automatically (release-tools auto-detects `doc/build_antora.sh`).

### Changes

- [x] **New Antora source tree** at `doc/modules/ROOT/` with 209 `.adoc` pages and 154 paired `.cpp`/`.txt` example sidecars.
- [x] **Wires release-tools' Antora pipeline** via `doc/build_antora.sh`, `doc/package.json` (antora 3.1.14 + lunr extension), and `doc/package-lock.json` for reproducible installs.
- [x] **Deletes 180 legacy HTML files** at `doc/*.html` / `doc/*.htm`, superseded by the Antora pages.
- [x] **Deletes legacy image assets** `doc/figs/` (106 files) and `doc/stoer_wagner_imgs/` (15 files); the new doc carries its own copies under `doc/modules/ROOT/assets/images/`.
- [x] **Deletes 3 partial reST remnants** (`read_graphml.rst`, `read_graphviz.rst`, `write_graphml.rst`), replaced by `doc/modules/ROOT/pages/io/{graphml,graphviz}.adoc`.
- [x] **Deletes `doc/Makefile`** (LaTeX/pdflatex/dvips rules from 2002, no longer used by any toolchain in the repo).

### Verification

- [x] `cd doc && ./build_antora.sh playbook.yml` succeeds end-to-end and produces `doc/build/site/graph/index.html`.

### Known follow-ups (not blocking this PR)

- I had to delete the develop louvain html documentation on this PR to solve conflicts: it will require a new `.adoc` documentation once this PR ships.
- CI freshness checks for examples output (`git diff --exit-code` on `.txt` after rebuild), may be best to defer until the random-generator examples are RNG-seeded so the diff is deterministic. Also maybe over-engineered.

Issue #457, opened 2026-02-16
undocumented filtered graph helpers?


andrea-cassioli-maersk:
Hi,
I stumbled across some helpers in `filter_graph.hpp` for which I cannot find any docs:

- `make_vertex_subset_filter`
- `make_vertex_subset_compliment_filter`

and the corresponding filters

- `is_in_subset`
- `is_not_in_subset`

Are they supposed to be exposed by the library?

Becheler:
Hi @andrea-cassioli-maersk  !

Thank you for this report 👍🏽 

For context, they're public convenience utilities located in `<boost/graph/filtered_graph.hpp>` :https://github.com/boostorg/graph/blob/develop/include/boost/graph/filtered_graph.hpp 

The git history shows that they were written by Jeremy Siek (the BGL author) in 2001, as part of the original filtered_graph implementation. They were always in the `boost::` namespace, never in `detail::` (so de facto considered public I guess). In 2010 Jeremiah Willcock found the typo `make_vertex_subset_compliment_filter` but explicitly preserved the old spelling for backwards compatibility and added a note saying to use the correct one `make_vertex_subset_compliment_filter`. I wonder at what point that could be removed (since we bumped to C++14 anyway).

Let's ask @jeremy-murphy  but I believe they are intentionally exposed helpers but undocumented because priority in early BGL was given to high-level algorithms and data structures, not trivial helpers like those.

andrea-cassioli-maersk:
Thank @Becheler .
I also noticed the filters only support ordered sets, possibly because unordered sets where not in the stl at that time. If we keep the functionalities, it would be nice to extend the supported type (and maybe have same filters for edges?).

Becheler:
You're right, it looks like it supports:
- `std::set`
- `boost::unordered_set`

But lacks support for:
- `std::unordered_set`
- `boost::unordered_flat_set`
- `boost::container_flat_set`
- generic containers with `.find()`, `.count()`, `.contains()`

Although I am not sure about the trade-off between code volume / utility for users. Unless I am mistaken, this is a one liner with modern C++14 lambdas:

```cpp
auto g2 = make_vertex_subset_filter(graph, set);
```
versus:
```cpp
auto g2 = filtered_graph(g, keep_all(), [&s](auto v){ return s.count(v); });
```

andrea-cassioli-maersk:
I agree, a lambda is enough! So maybe the question is whether they should either documented, or removed altogether (not sure if deprecation is a thing).

[...]

PR #502, opened 2026-06-07
Personalized page rank

maniospas:
### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

Provides an implementation of personalized PageRank following the motivating issue's outline for the provided interface.

### Motivation

Refs #493 

### Testing

`personalized_pagerank_test.cpp` was added to tests. This runs two node scoring scenarios on a graph with well-understood structure, where assertions are added for equality between isomorphic nodes and certain interesting numerical comparisons checked to hold. The build command for local testing was `g++ -std=c++14 -I./boost -o test test.cpp -O2`.

### Checklist

Documentation change not applicable, but likely new documentation should be added to reflect the introduced algorithm.

### DIFF
--- a/include/boost/graph/personalized_page_rank.hpp
+++ b/include/boost/graph/personalized_page_rank.hpp
@@ -185,7 +185,7 @@ namespace graph
                 }
                 l1_norm += l1_accumulated_norm;
             }
-            BGL_FORALL_VERTICES_T(v, g, Graph) put(to_rank, v, get(to_rank, v)/l1_norm);
+            for (auto v : boost::make_iterator_range(vertices(g))) put(to_rank, v, get(to_rank, v)/l1_norm);
         }
 
         template <
@@ -194,7 +194,7 @@ namespace graph
             typename PersonalizationMap,
             typename RankMap,
             typename RankMap2 >
-        void page_rank_step(
+        void personalized_page_rank_step(
             const Graph& g,
             WeightMap weight_map,
             PersonalizationMap personalization_map,
@@ -205,15 +205,16 @@ namespace graph
         {
             typedef typename property_traits< RankMap >::value_type damping_type;
             damping_type l1_norm(0); // Computing the norm simultaneously avoids an extra summing iteration.
-            BGL_FORALL_VERTICES_T(v, g, Graph)
+            for (auto v : boost::make_iterator_range(vertices(g)))
             {
                 typename property_traits< RankMap >::value_type rank(0);
-                BGL_FORALL_INEDGES_T(v, e, g, Graph) rank += get(from_rank, source(e, g))*get(weight_map, e);//get(from_rank, source(e, g)) / out_degree(source(e, g), g);
+                for (auto e : boost::make_iterator_range(in_edges(v, g))) 
+                    rank += get(from_rank, source(e, g))*get(weight_map, e);
                 auto v_score = (damping_type(1) - damping) * get(personalization_map, v) + damping * rank;
                 put(to_rank, v, v_score);
                 l1_norm += v_score;
             }
-            BGL_FORALL_VERTICES_T(v, g, Graph) put(to_rank, v, get(to_rank, v)/l1_norm);
+            for (auto v : boost::make_iterator_range(vertices(g))) put(to_rank, v, get(to_rank, v)/l1_norm);
         }
     } // end namespace detail
 
@@ -238,12 +239,12 @@ namespace graph
         typedef typename property_traits< PersonalizationMap >::value_type rank_type;
 
         rank_type personalization_norm(0);
-        BGL_FORALL_VERTICES_T(v, g, Graph) personalization_norm += get(personalization_map, v);
+        for (auto v : boost::make_iterator_range(vertices(g))) personalization_norm += get(personalization_map, v);
 
         // TBD: This implementation couples iterators when possible under reduced L1 cache invalidation assumptions,
         // but this is not necessarily the case because we may be grabbing 2x memory lanes each time to write there.
         // Should investigate which pattern is faster.
-        BGL_FORALL_VERTICES_T(v, g, Graph)
+        for (auto v : boost::make_iterator_range(vertices(g)))
         {
             rank_type value = get(personalization_map, v)/personalization_norm;
             put(personalization_map, v, value);
--- a/include/boost/graph/personalized_page_rank.hpp
+++ b/include/boost/graph/personalized_page_rank.hpp
@@ -22,125 +22,23 @@ namespace boost
 {
 namespace graph
 {
-    namespace personalized_page_rank_detail
+    struct rank_convergence
     {
-        template < typename Graph>
-        std::size_t in_or_out_degree(
-            typename boost::graph_traits<Graph>::vertex_descriptor v,
-            const Graph& g,
-            bidirectional_graph_tag)
-        {
-            return in_degree(v, g);
-        }
-        template < typename Graph>
-        std::size_t in_or_out_degree(
-            typename boost::graph_traits<Graph>::vertex_descriptor v,
-            const Graph& g,
-            incidence_graph_tag)
-        {
-            return out_degree(v, g);
-        }
-    }
-
-    template <typename Graph>
-    boost::function_property_map<
-        std::function<double(typename boost::graph_traits<Graph>::edge_descriptor)>,
-        typename boost::graph_traits<Graph>::edge_descriptor>
-    markovian_weights(const Graph& g)
-    {
-        using Edge = typename boost::graph_traits<Graph>::edge_descriptor;
-        using Func = std::function<double(Edge)>;
-        Func f = [&g](Edge e)
-        {
-            auto u = source(e, g);
-            std::size_t denom;
-            denom = out_degree(u, g) ;
-            return 1.0 / denom;
-        };
-        return make_function_property_map<Edge, double>(f);
-    }
-
-    template <typename Graph>
-    boost::function_property_map<
-        std::function<double(typename boost::graph_traits<Graph>::edge_descriptor)>,
-        typename boost::graph_traits<Graph>::edge_descriptor>
-    spectral_weights(const Graph& g)
-    {
-        using Edge = typename boost::graph_traits<Graph>::edge_descriptor;
-        using Func = std::function<double(Edge)>;
-        Func f = [&g](Edge e)
-        {
-            typedef typename boost::graph_traits<Graph>::traversal_category category;
-            auto u = source(e, g);
-            auto v = target(e, g);
-            std::size_t denom;
-            denom = out_degree(u, g) * personalized_page_rank_detail::in_or_out_degree(v, g, category());
-            return 1.0 / std::sqrt(denom);
-        };
-        return make_function_property_map<Edge, double>(f);
-    }
-
-    template <typename Graph>
-    boost::function_property_map<
-        std::function<double(typename boost::graph_traits<Graph>::edge_descriptor)>,
-        typename boost::graph_traits<Graph>::edge_descriptor>
-    renormalized_weights(const Graph& g)
-    {
-        using Edge = typename boost::graph_traits<Graph>::edge_descriptor;
-        using Func = std::function<double(Edge)>;
-        Func f = [&g](Edge e)
-        {
-            typedef typename boost::graph_traits<Graph>::traversal_category category;
-            auto u = source(e, g);
-            auto v = target(e, g);
-            std::size_t denom;
-            denom = (1+out_degree(u, g)) * (1+personalized_page_rank_detail::in_or_out_degree(v, g, category()));
-            return 1.0 / std::sqrt(denom);
-        };
-        return make_function_property_map<Edge, double>(f);
-    }
-
-    struct vertex_map_convergence
-    {
-        explicit vertex_map_convergence(std::size_t n, double tol=0) : n(n), tol(tol) {} // allowing tolerance for early stopping
-
+        explicit rank_convergence(std::size_t iters, double tol=0) : iters(iters), tol(tol) {} // allowing tolerance for early stopping
         template < typename RankMap, typename RankMap2, typename Graph >
-        bool operator()(
-            const RankMap& current,
-            const RankMap2& previous,
-            const Graph& g)
+        bool operator()(const RankMap& current, const RankMap2& previous, const Graph& g)
         {
-            if ( n-- == 0 )
-            {
+            if (--iters == 0)
                 return true;
-            }
             if (!tol)
-            {
                 return false;
-            }
-
-            typedef typename property_traits< RankMap >::value_type rank_type;
+            using rank_type = typename property_traits< RankMap >::value_type;
             rank_type sum_abs(0);
             for (auto v : boost::make_iterator_range(vertices(g)))
-            {
-                rank_type difference = get(current, v)-get(previous, v);
-                sum_abs += std::abs(difference);
-            }
+                sum_abs += std::abs(get(current, v) - get(previous, v));
             return sum_abs*num_vertices(g)<tol;
         }
-
-        std::size_t get_remaining_iters() const
-        {
-            return n;
-        }
-
-        bool has_converged() const 
-        {
-            return n || !tol;
-        }
-
-    private:
-        std::size_t n;
+        std::size_t iters;
         double tol;
     };
 
@@ -161,17 +59,18 @@ namespace graph
             typename property_traits< RankMap >::value_type damping,
             incidence_graph_tag)
         {
-            typedef typename property_traits< RankMap >::value_type rank_type;
+            using rank_type = typename property_traits< RankMap >::value_type;
             rank_type l1_norm(0); // Computing the norm simultaneously avoids an extra summing iteration.
 
-            // Initialize the constant part of maps
+            // Initialize the constant part of maps.
             for (auto v : boost::make_iterator_range(vertices(g))) 
             {
                 auto v_constant = rank_type(1 - damping) * get(personalization_map, v);
                 put(to_rank, v, v_constant);
                 l1_norm += v_constant;
             }
-
+            
+            // Maintenance comment:
             for (auto u : boost::make_iterator_range(vertices(g)))
             {
                 rank_type u_rank_factor = damping * get(from_rank, u);
@@ -185,7 +84,11 @@ namespace graph
                 }
                 l1_norm += l1_accumulated_norm;
             }
-            for (auto v : boost::make_iterator_range(vertices(g))) put(to_rank, v, get(to_rank, v)/l1_norm);
+            // If there are negative edge weights, or if negative damping is used, l1_norm could be zero or near-zero. 
+            // Division in those cases is conceptually correct for floating point weights, and actually expected behavior. 
+            // That said, such edge cases are impossible to arise for all typical algorithm uses.
+            for (auto v : boost::make_iterator_range(vertices(g))) 
+                put(to_rank, v, get(to_rank, v)/l1_norm); 
         }
 
         template <
@@ -203,20 +106,22 @@ namespace graph
             typename property_traits< RankMap >::value_type damping,
             bidirectional_graph_tag)
         {
-            typedef typename property_traits< RankMap >::value_type damping_type;
+            using damping_type = typename property_traits< RankMap >::value_type;
             damping_type l1_norm(0); // Computing the norm simultaneously avoids an extra summing iteration.
             for (auto v : boost::make_iterator_range(vertices(g)))
             {
-                typename property_traits< RankMap >::value_type rank(0);
+                damping_type rank(0);
                 for (auto e : boost::make_iterator_range(in_edges(v, g))) 
                     rank += get(from_rank, source(e, g))*get(weight_map, e);
                 auto v_score = (damping_type(1) - damping) * get(personalization_map, v) + damping * rank;
                 put(to_rank, v, v_score);
                 l1_norm += v_score;
             }
-            for (auto v : boost::make_iterator_range(vertices(g))) put(to_rank, v, get(to_rank, v)/l1_norm);
+            // See above function for potential division by zero comments.
+            for (auto v : boost::make_iterator_range(vertices(g))) 
+                put(to_rank, v, get(to_rank, v)/l1_norm);
         }
-    } // end namespace detail
+    } // end namespace personalized_page_rank_detail
 
     template <
         typename Graph,
@@ -232,25 +137,24 @@ namespace graph
         RankMap rank_map,
         Done done,
         typename property_traits< RankMap >::value_type damping,
-        typename graph_traits< Graph >::vertices_size_type n,
         RankMap2 rank_map2
         BOOST_GRAPH_ENABLE_IF_MODELS_PARM(Graph, vertex_list_graph_tag))
     {
-        typedef typename property_traits< PersonalizationMap >::value_type rank_type;
-
+        using rank_type = typename property_traits< PersonalizationMap >::value_type;
         rank_type personalization_norm(0);
-        for (auto v : boost::make_iterator_range(vertices(g))) personalization_norm += get(personalization_map, v);
+        for (auto v : boost::make_iterator_range(vertices(g))) 
+            personalization_norm += get(personalization_map, v);
 
         // TBD: This implementation couples iterators when possible under reduced L1 cache invalidation assumptions,
         // but this is not necessarily the case because we may be grabbing 2x memory lanes each time to write there.
-        // Should investigate which pattern is faster.
+        // Could investigate which pattern is faster in the future.
         for (auto v : boost::make_iterator_range(vertices(g)))
         {
             rank_type value = get(personalization_map, v)/personalization_norm;
             put(personalization_map, v, value);
             put(rank_map, v, value);
         }
-
+        
         bool to_map_2 = true;
         do
         {
@@ -260,10 +164,12 @@ namespace graph
             else
                 personalized_page_rank_detail::personalized_page_rank_step(g, weight_map, personalization_map, rank_map2, rank_map, damping, category());
             to_map_2 = !to_map_2;
-        } while ((to_map_2  && !done(rank_map,  rank_map2, g)) || (!to_map_2 && !done(rank_map2, rank_map,  g)));
+        } 
+        while ((to_map_2  && !done(rank_map,  rank_map2, g)) || (!to_map_2 && !done(rank_map2, rank_map,  g))); // Done may not be symmetric.
         
         // Now multiply the result with personalization_norm to restore the order of magnitude and store it in rank_map.
-        // Also restore the original personalization_map's magnitude for reuse (this is lossy up to numerical tolerance but leaner).'
+        // Also restore the original personalization_map's magnitude for reuse (this is lossy up to numerical tolerance 
+        // but leaner than making a copy).
         if (!to_map_2)
         {
             for (auto v : boost::make_iterator_range(vertices(g)))
@@ -295,61 +201,33 @@ namespace graph
         PersonalizationMap personalization_map,
         RankMap rank_map,
         Done done,
-        typename property_traits< RankMap >::value_type damping,
-        typename graph_traits< Graph >::vertices_size_type n)
+        typename property_traits< RankMap >::value_type damping)
     {
-        typedef typename property_traits< RankMap >::value_type rank_type;
-
+        using rank_type = typename property_traits< RankMap >::value_type;
         std::vector< rank_type > ranks2(num_vertices(g));
-        return personalized_page_rank(g, weight_map, personalization_map, rank_map, done, damping, n,
+        return personalized_page_rank(g, weight_map, personalization_map, rank_map, done, damping,
             make_iterator_property_map(ranks2.begin(), get(vertex_index, g)));
     }
 
-    template <
-        typename Graph,
-        typename WeightMap,
-        typename PersonalizationMap,
-        typename RankMap,
-        typename Done >
-    inline Done personalized_page_rank(
-        const Graph& g,
-        WeightMap weight_map,
-        PersonalizationMap personalization_map,
-        RankMap rank_map,
-        Done done,
-        typename property_traits< RankMap >::value_type damping = 0.85)
-    {
-        return personalized_page_rank(g, weight_map, personalization_map, rank_map, done, damping, num_vertices(g));
-    }
-
-    template <
-        typename Graph,
-        typename WeightMap,
-        typename PersonalizationMap,
-        typename RankMap >
-    inline vertex_map_convergence personalized_page_rank(
-        const Graph& g,
-        WeightMap weight_map,
-        PersonalizationMap personalization_map,
-        RankMap rank_map,
-        typename property_traits< RankMap >::value_type damping = 0.85)
-    {
-        std::size_t n_iters(1.0/(1-damping)+0.5); // accounts for "most" random walks
-        auto convergence = vertex_map_convergence(n_iters);
-        return personalized_page_rank(g, weight_map, personalization_map, rank_map, convergence, damping, num_vertices(g));
-    }
-
-    template <
-        typename Graph,
-        typename PersonalizationMap,
-        typename RankMap >
-    inline vertex_map_convergence personalized_page_rank(
+    template < typename Graph, typename PersonalizationMap, typename RankMap >
+    rank_convergence personalized_page_rank(
         const Graph& g,
         PersonalizationMap personalization_map,
         RankMap rank_map,
-        typename property_traits< RankMap >::value_type damping = 0.85)
+        typename property_traits< RankMap >::value_type damping=0.85)
     {
-        return personalized_page_rank(g, asymmetric_normalization(g), personalization_map, rank_map);
+        // This is the most traditional personalized PageRank implementation, with minimized signature.
+        using Edge = graph_traits<Graph>::edge_descriptor;
+        using rank_type = typename property_traits< RankMap >::value_type;
+        std::vector< rank_type > ranks2(num_vertices(g));
+        auto markovian_weights = make_function_property_map<Edge, double>([&g](Edge e){ return 1.0 / out_degree(source(e, g), g); });
+        return personalized_page_rank(g,
+            markovian_weights, 
+            personalization_map, 
+            rank_map, 
+            rank_convergence(100, 1.E-9), 
+            damping,
+            make_iterator_property_map(ranks2.begin(), get(vertex_index, g)));
     }
 
 }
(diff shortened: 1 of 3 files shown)


include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,358 @@
...
+        {
+            typedef typename boost::graph_traits<Graph>::traversal_category category;
+            auto u = source(e, g);
+            auto v = target(e, g);
+            std::size_t denom;
+            denom = (1+out_degree(u, g)) * (1+personalized_page_rank_detail::in_or_out_degree(v, g, category()));
+            return 1.0 / std::sqrt(denom);
+        };
+        return make_function_property_map<Edge, double>(f);
+    }
```
Becheler: I'm not a huge fan of those lines.

2. Generic names in the `shared boost::graph` namespace: `spectral_weights` / `renormalized_weights` are terms other algorithms (spectral clustering, spectral layout…) could plausibly want too: algorithm-specific helpers shouldn't claim them.
3. They're free functions taking a `Graph`, so they're pulled in by argument-dependent lookup, that sounds risky for little benefit.
4. spectral/renormalized silently fall back to out-degree on directed graphs without in-edges, changing the normalization with no error or warning diagnostic
4. Its ~75 lines of code to review, test and maintain when they can be collapsed to one-liners by the user/caller:

```cpp
using Edge = graph_traits<Graph>::edge_descriptor;

//// Pick one:

// markovian
auto weights = make_function_property_map<Edge, double>(
    [&g](Edge e){ return 1.0 / out_degree(source(e, g), g); });

// spectral
auto weights = make_function_property_map<Edge, double>(
    [&g](Edge e){ return 1.0 / std::sqrt(double(out_degree(source(e, g), g) * out_degree(target(e, g), g))); });

// renormalized
auto weights = make_function_property_map<Edge, double>(
    [&g](Edge e){ return 1.0 / std::sqrt(double((1 + out_degree(source(e, g), g)) * (1 + out_degree(target(e, g), g)))); });

// user just injects their pick:
personalized_page_rank(g, weights, personalization, rank);
```
I understand the will to test and offer those easy lines to users, but I think this is much more the responsibility of the unit test + a good `example.cpp` + the documentation page to offer a bunch of possible applications 😄

For example, look at https://en.cppreference.com/cpp/algorithm/sort : the documentation (not the library!) offers a 3 different variations on how to inject an arbitrary comparator.
 
What do you think ?

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,358 @@
...
+
+        bool has_converged() const 
+        {
+            return n || !tol;
+        }
+
+    private:
+        std::size_t n;
+        double tol;
+    };
```
Becheler: There is a bit of over-promises here from the library standpoint:
- the naming `boost::graph::vertex_map_convergence` has even higher collisions risks since it overlaps any algorithm related to measuring convergence of metrics over vertices (label propagation, diffusion...) : please consider renaming it e.g. `rank_convergence`
- `has_converged` and `get_remaining_iters` are reporting utilities, not core utilities. They don't help illustrating the contract (the algorithm only needs `operator()()` and they increase the API/documentation surface. I would rather ship the most basic type, and show in the tests/doc/examples how the user can extend it via inheritance:

```cpp
// library responsibility: minimal criterion
struct rank_convergence {
    explicit rank_convergence(std::size_t max_iters, double tol = 0);
    // operator()(current, previous, g): if (max_iters == 0) return true; --max_iters; ...
protected:
    std::size_t max_iters;
    double tol;
};

// caller in tests/examples/docs/real-life can opt-in reporting easily:
struct reporting_convergence : rank_convergence {
    using rank_convergence::rank_convergence; // inherit constructors
    bool has_converged() const { return max_iters != 0; }
    std::size_t remaining() const { return max_iters; }
};
```

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,358 @@
...
+        BOOST_GRAPH_ENABLE_IF_MODELS_PARM(Graph, vertex_list_graph_tag))
+    {
+        typedef typename property_traits< PersonalizationMap >::value_type rank_type;
+
+        rank_type personalization_norm(0);
+        for (auto v : boost::make_iterator_range(vertices(g))) personalization_norm += get(personalization_map, v);
+
+        // TBD: This implementation couples iterators when possible under reduced L1 cache invalidation assumptions,
+        // but this is not necessarily the case because we may be grabbing 2x memory lanes each time to write there.
+        // Should investigate which pattern is faster.
```
Becheler: If you are not sure the effect is real and don't have the benchmark harness to measure it, it's completely ok: we should favor whichever implemention is simplest and iterate on performance in future PRs.

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,358 @@
...
+    Done personalized_page_rank(
+        const Graph& g,
+        WeightMap weight_map,
+        PersonalizationMap personalization_map,
+        RankMap rank_map,
+        Done done,
+        typename property_traits< RankMap >::value_type damping,
+        typename graph_traits< Graph >::vertices_size_type n)
+    {
+        typedef typename property_traits< RankMap >::value_type rank_type;
```
Becheler: please prefer using the modern syntax `using alias_type = aliased_type` 😄

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,358 @@
...
+        typename RankMap,
+        typename Done >
+    Done personalized_page_rank(
+        const Graph& g,
+        WeightMap weight_map,
+        PersonalizationMap personalization_map,
+        RankMap rank_map,
+        Done done,
+        typename property_traits< RankMap >::value_type damping,
+        typename graph_traits< Graph >::vertices_size_type n)
```
Becheler: I believe this `n` is completely unused throughout the code ?
looks like in pageRank, `n` was used as uniform distribution, and PPR does not need it anymore ?

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,358 @@
...
+        typename PersonalizationMap,
+        typename RankMap >
+    inline vertex_map_convergence personalized_page_rank(
+        const Graph& g,
+        WeightMap weight_map,
+        PersonalizationMap personalization_map,
+        RankMap rank_map,
+        typename property_traits< RankMap >::value_type damping = 0.85)
+    {
+        std::size_t n_iters(1.0/(1-damping)+0.5); // accounts for "most" random walks
```
Becheler: I would document this in a precondition because as is, this is UB if e.g. damping = 1.0, which is a plausible user input
```cpp
BOOST_ASSERT_MSG(damping >= 0.0 && damping < 1.0, "personalized_page_rank: damping must be in [0, 1)");
```

Becheler (changes requested):
Thanks again for so much work ! 🙏🏽 
I think we can reach a leaner PR with simpler API surface by moving some things from the library to the user. It has greater generality and does not overwhelm the library code.

Let me know what you think, 🚀 
Arno

maniospas:
Hello again both and thank you for looking through the PR. :smile: 

All the points @Becheler raises make a lot of sense. So, if I understand correctly, algorithms should refrain from adding domain knowledge in the form of DSL-like accompaniments, as these can just be presented in the documentation. Did I get this right?

If my understanding is correct, then perhaps all overloads that automatically construct a convergence manager should also be removed, as those versions use domain-backed rules. In that case, the convergence class can be removed altogether too. Thus, after removing everything else other the iteration helpers (as well as `n` from arguments, which in retrospect is not needed in the modern interface -  it was a leftover from using the original `page_rank` as a prototype), the only signatures that would exist for the algorithm could be the much leaner:

```c++
 Done personalized_page_rank(g, weight_map, personalization_map, rank_map, done, damping, rank_map2);
 Done personalized_page_rank(g, weight_map, personalization_map, rank_map, done, damping);
```

There could be an argument to, for historical reasons (mainly evoking the idea of non-personalized PageRank) also have the next signature for Markovian normalization while keeping the `rank_convergence` and setting it to run for 100 iters with 1.E-9 tolerance. However, this contradicts the above principle of avoiding hardcoding domain knowledge in the library. Thoughts?

```c++
 Done personalized_page_rank(g, personalization_map, rank_map, damping=0.85);
```

Becheler:
You're welcome @maniospas  ! 🤗 

Let's say it this way: for a first PR in uncertain weather (BGL has more steam power these days, but many directions are left uncertain/unexplored) we should take minimal risk and avoid baking opinionated domain policy into the API (it makes your code more powerful, not less!).

1. Weight normalizations (markovian/spectral/renormalized) are domain policy. Agreed: drop from the API, show them as one-liners in the example/docs until BGL directions get cleared:
    - There **are** discussions about wether such things deserve names in BGL or not (see e.g. #457 , you are welcome to contribute/chime in). We don't know under what form yet (lambdas, structures, namespaces, etc...)
    - There are also discussions about making property maps composition easier (see #483 ): if it lands, the idiomatic form might become `g >> edge_weight >> your_markovian >> w`, where the transform is just a user callable plugged into generic composition. We don't yet know if these normalizations get names at all, or in what shape (lambda / struct / CPO / namespaces ). That's exactly why I'd rather not freeze one form into the API today.
    - Those two discussions and your current PR are conceptually linked, but their interaction quite unclear yet. I just don't want you to be caught in the crossfire: your algorithm contribution should remain relevant whatever the answer to this side-quest end up being, and the features you offer to users in 2026 should not be deprecated in 2027. It will always be time to add things once the directions are settled ⛵ 
2. Yes I think the two overloads you showed are the minimal viable product 😉 
3. That being said, you are the PageRank expert (I'm just the lib guy lol). So if you tell me that 80% of users will want to reach for `Done personalized_page_rank(g, personalization_map, rank_map, damping=0.85);` then it's prior information worth merging. But in that case I would define what this specific overload needs internally, and not expose (yet!) anything to its users. 
4. I also think that the minimal convergence struct (without reporting) is a special case: it formalize the minimal viable interface/semantics a convergence structure must have to make your algorithm run. It's also mirroring the PageRank way. So I'd say this one can stay with us ? But I see your point, it's a gut feeling more than a hardcore guideline...

Does it make sense ?

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+    template < typename Graph, typename PersonalizationMap, typename RankMap >
+    rank_convergence personalized_page_rank(
+        const Graph& g,
+        PersonalizationMap personalization_map,
+        RankMap rank_map,
+        typename property_traits< RankMap >::value_type damping=0.85)
+    {
+        // This is the most traditional personalized PageRank implementation, with minimized signature.
+        using Edge = graph_traits<Graph>::edge_descriptor;
```
Becheler: you need a typename here too ;)

```
./../../boost/graph/personalized_page_rank.hpp:220:22: error: need ‘typename’ before ‘boost::graph_traits<Graph>::edge_descriptor’ because ‘boost::graph_traits<Graph>’ is a dependent scope
  220 |         using Edge = graph_traits<Graph>::edge_descriptor;
      |                      ^~~~~~~~~~~~~~~~~~~
      |                      typename 
```

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,236 @@
...
+        const Graph& g,
+        PersonalizationMap personalization_map,
+        RankMap rank_map,
+        typename property_traits< RankMap >::value_type damping=0.85)
+    {
+        // This is the most traditional personalized PageRank implementation, with minimized signature.
+        using Edge = typename graph_traits<Graph>::edge_descriptor;
+        using rank_type = typename property_traits< RankMap >::value_type;
+        std::vector< rank_type > ranks2(num_vertices(g));
+        auto markovian_weights = make_function_property_map<Edge, double>([&g](Edge e){ return 1.0 / out_degree(source(e, g), g); });
```
Becheler, line 239: perfect ! 😄

test/personalized_pagerank_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+    personalization[3] = 1;
+
+    std::size_t max_iters(100); // Convergence is so bad in this graph that it needs such a high cap.
+    using Edge = graph_traits<Graph>::edge_descriptor;
+    auto weight = make_function_property_map<Edge, double>(
+        [&g](Edge e){ return 1.0 / std::sqrt(double(out_degree(source(e, g), g) * out_degree(target(e, g), g))); });
+    auto convergence1 = graph::rank_convergence(max_iters, 1.E-9);
+    convergence1 = graph::personalized_page_rank(g, weight, personalization_map, rank_map, convergence1, 0.9);
+
+    std::cout << "ended after "<<max_iters-convergence1.iters << " iterations\n";
```
Becheler: For last round of review: please refrain from using standard output in tests. Standard outputs are brittle (can't be automatically checked in CI) and make poor non-regression tools.
If the output is predictible, then use assertions: breaking an assertion will stop the CI, breaking an output will not break the CI 😉

Becheler: That being said, for a later documentation PR, the `examples.cpp` scripts can completely use standard output and assertions, as the output is run in CI and integrated into the doc: see https://491.graph.prtest3.cppalliance.org/graph/algorithms/metrics/page_rank.html
But unit tests and examples serve different purpose: 
- unit tests should always fail if something expected is broken
- unit tests should use a dedicated logger if something can't be expected/guaranteed (that's one more discussion we have on BGL rn)
- examples should use (static_)assert to simplify reading (it's easier to read `assert(expected == computed)` than reading the output to visually compare the two
- examples should use output for brief, visual proofs/illustrations: it may be worth relying on visual checks if a formal check would double the size of the example script.

PS: yes this should go to the contribution guidelines, I just added an appendix section about it in #495 : thank you 🙏🏽

test/personalized_pagerank_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+    BOOST_ASSERT(ranks[0]<ranks[1]+0.1);
+    BOOST_ASSERT(ranks[8]==ranks[9]);
+    BOOST_ASSERT(ranks[11]==ranks[12]);
+    BOOST_ASSERT(ranks[11]>0);
+    BOOST_ASSERT(convergence1.iters<max_iters);
+
+    // COMPUTING A HIGH-PASS VERSION BY PASSING NEGATIVE DAMPING (resulting values can be negative)
+    // This is not completely correct yet for computing graph gradients, because it needs to eskew normalization,
+    // so we need to be able to customize said normalization.
+    // Damping should generaly be in the range [-1,1].
```
Becheler: What do you mean ? Is the current implementation broken without said normalization customization point, or is it just that the test is a bit unrealistic yet proving correctness ?

Becheler:
Hi @maniospas !

Thanks for editing the files, I think it's perfect like this 🏆 

For the last step, it should be pretty straightforward now: just scan what the algorithm requires on the templated types (graph concepts, property map concept), expand a bit the test if you can think of corner edges (damping factor definition domain violation, different graph types, different property maps). You don't have to go full 500 loc test file, but maybe defining two test functions outside the main and calling them with e.g adjacency list, adjacency matrix, csr, would already strengthen your implementation and guarantee it works across a reasonably variable range of data structs, something like:

```cpp

template <typename Graph>
void test_basic(const Graph& g)
{
    // basically what you already have I believe
}

template <typename Graph>
void test_corner_cases(const Graph& g)
{
  // just add a few corner cases you can think of ? If any, no worries
}

int main()
{
    typedef adjacency_list<vecS, vecS, directedS> AdjList;
    typedef adjacency_matrix<directedS> AdjMatrix;
    typedef compressed_sparse_row_graph<directedS> CSR;

    // build the same small graph for each backend, then run both suites
    {
        AdjList g; /* add_edge(...) */
        test_basic(g);
        test_corner_cases(g);
    }
    {
        AdjMatrix g(/* n */); /* add_edge(...) */
        test_basic(g);
        test_corner_cases(g);
    }
    {
        CSR g(/* edges_are_unsorted, begin, end, n */);
        test_basic(g);
        test_corner_cases(g);
    }

    return boost::report_errors();
}
```

Intuitively I would think doubling your test file size would be ok, tripling it would be overdoing it (just for an order of magnitude) 😄

maniospas:
Hello @Becheler again and many thanks for the feedback and helping me with this PR. :smile: 

I am pretty happy where the whole thing ended up, too. I added the last overload because it is ~80% what people new to node ranking will want to use, though it is sufficient for perhaps only ~20% of end-products (numbers not real: trying to convey my intuition here, in case it is unwanted). 

By the way, I took the liberty of having the remaining iterations and tolerance to be public struct fields to allow greater versatility without adding any class methods for convergence. I am not sure if you noticed and are ok with it, so mentioning it here. 

With regards to asserts, I can see even negative damping and negative edge weights to be viable options, so I would ideally not have any numerical restrictions. Even zero normalization yielding NaNs could be meaningful information, in my opinion. At least, for me, it feels that if there is no error convergence there also should be no error for more "exotic" values that still make sense. Perhaps a good compromise assert guard would be for dampening to be in the range [-1,1] (the negative values are needed to be able to express high-pass gradients via the Laplacian) but I am not sure if adding such a restriction is meaningful. Is there a chance the algorithm can expose warnings instead of errors? Perhaps this could be the purview of feature visitors once the respective discussion is stabilized.

Working on the concept asserts and tests now. Will ping you again for the result, because I am not confident about my understanding of available concepts. For tests, I will add the breadth you mention - there should be no problem staying in a small volume while creating a matrix of option combinations.

A final question I have is whether accompanying documentation should be part of this PR or be provided separately?

test/personalized_pagerank_test.cpp
maniospas: The current implementation is not 100% a match of the expected Laplacian computations, because we would need to be able to control the normalization inside the algorithm (currently L1 normalization inside should become no normalization). Not passing a custom normalizer was one of our simplification points, because the proper generalization would be an optimizer. 

Yielding negative weights is expected behavior, but the computations may not be 100% correct (though I do not remember the reason I thought so right now). I will take a look again to see when lack of correctness occurs, because the result may be provably equivalent even if the computational schema is not what is typically seen. 

That said, this particular usage is so rare, that I believe it should not be an issue for an MVP. The test runs mainly to demonstrate that there is no issue handling inputs that encroach into negative scoring territory.

Becheler: Ok then it's perfectly in line with our previous conversations, thanks so much for taking the time to re-explain it to me :)

Becheler:
> I added the last overload because it is ~80% what people new to node ranking will want to use, though it is sufficient for perhaps only ~20% of end-products

Perfect, it seems amazing for a first step ! 👍🏽 

> I took the liberty of having the remaining iterations and tolerance to be public struct fields to allow greater versatility 

Ooh this escaped my attention, thank for mentioning it. From a self-documenting perspective, I think we want them to remain `private` or `protected`. By exposing everything in public you end up with no automatic way to guarantee your algorithm never touches those internals (and does not even need them, not even their type). So users would get confused. From a maintainance perspective, it means: expect a "does my predicate need a `iters` and `tol` member?" issue to be filed in a few months/years, requiring to dive back into the code. And then you would have to write down the whole thing in the doc to clear up misconceptions:

> `although made public in the code of the predicate, the predicate data members `iters` and `tol` are not directly used by the algorithm internals, so users do not need them as the `operator()(...)` logic is sufficient, for example for emulating infinite iterations driven only by tolerance.

Stuff like that. It's lot of extra words to explain that the only public contract the user should fulfill is to expose only:
```cpp
bool operator()(const RankMap& current, const RankMap2& previous, const Graph& g)
```
and nothing more. You can mark them `private` to make it closed to extension (users need to copy-paste your class to modify it) or `protected` (users can extend the struct behavior through inheritance). Just my 2cents 😉 

> I would ideally not have any numerical restrictions.

I will definitely trust your expert knowledge here !

> because I am not confident about my understanding of available concepts.

Nobody is, so don't hesitate asking questions here or in the discussions ! Just know you can probably finish writing the tests and the code without concepts, those come at the very very end to guarantee you don't run into wrong operations because you passed e.g. a property map that can only be read (e.g. a lambda returning 42 for every vertex: ReadablePropertyMap) when you also need to write to it (ReadWritePropertyMap). Concepts will not change your code behavior or make tests fail: they will just give cleaner compile error messages if users did something wrong so that would be last thing I'd write.

> A final question I have is whether accompanying documentation should be part of this PR or be provided separately?

Another cross-fire I don't want you to get caught into 😆 
We are in the middle of the decision to merge PR #491 : this means currently you have no proper way to write it:
- writing it the old html way could be deprecated next week
- the documentation infrastructure required for you to write the documentation and example does not yet exist in `develop`
- so yes, probably a next PR would be the path of less friction for you 🙏🏽

maniospas:
Hello @Becheler and apologies for the delay; I had some personal issues. Letting you know that I will now be resuming work on this PR assuming the previous status. Nonetheless, if some things have changed or been stabilized in the interim (e.g., doc format) I would be happy to adjust/include them.

Becheler:
Hi @maniospas ! 😄 

I was considering asking you if everything was ok ! I am glad you are back and I hope you are doing well.
No worries at all for the delay of course (and thanks for coming back ❤️  )

There were positives to this delay, mainly that the documentation PR was merged, so writing the documentation page is now easier than ever! But that will be the last step so you don't have to worry about it yet !

Welcome back, happy to help pick things up if you need me to  🎉

maniospas:
Hello again and thanks for the kind of words. :-) 

Tried to add concept asserts in the main algorithm, if you can check them, as well as a first matrix of directed graph tests (will add directed tests too to compare normal PageRank with a bidirectional run). Not sure at all if I checked against the correct concepts.

Also added an assert on the damping after all, because it was revealed during edge case testing that certain unstable divergent computations somehow do not preserve symmetry, whereas I was expecting the computations to create the overflows but mathematically preserve subgraph isomorphisms even in edge cases.

Tangent: Not certain why exactly the isomoprhism property is not preserved; I thought the computations would at worst become infinity per IEEE754 and then nan, but perhaps there's a numerical underflow in a denominator. Thus, I think the likely "fix" would be to traverse nodes in rank order but this creates `O(V(log V + E) )` runtime due to sorting and I'd rather not go into a rabbit hole given that these values have not yet found any theoretical or practical application. Perhaps there could be a kind of rankmap for the results that maintains sorted node order during iterations and thus allow the assert to conditionally remove the bounds, but this is likely work for the future.

P.S. Random feedback on navigating the documentation on concepts: I think I am kind of missing a place to see all concepts side-by-side/list/dependency graph (whatever makes sense) linked from all concept pages, because their descriptions are really great at explaining differences but it was kind of burdensome to keep switching pages to grok the details.

Becheler:
I will have a look thank you !

In the meantime, would you be able to sync your fork with the boost repo, then rebase your PR branch on develop so the PR shows only your commits ? I don't know what happened but suddenly 170 commits and 959 modified files were dropped on the PR and it's hard to track what changed :) 

Thank you !

Edit: I don't know if it's too late to do so. It looks like develop was merged into the PR branch rather than the PR branch rebased on top of develop :) Also make a local copy before doing e.g. `a git push --force` ;)

Edit2: IIUC it's too late for a clean rebase. I think you would want something like this: create a new branch from develop, move to it your own commits, then repoint to the PR branch

```
# upstream = boostorg/graph, origin = your fork
git fetch upstream

# See ONLY your real work (excludes merges and develop's commits):
git log --no-merges upstream/develop..personalized_page_rank --oneline

# Fresh branch off current develop:
git checkout -b pagerank-clean upstream/develop

# Cherry-pick just those SHAs commits, oldest first:
git cherry-pick <sha_oldest> ... <sha_newest>

# Repoint to the PR branch:
git push --force-with-lease origin pagerank-clean:personalized_page_rank
```

Of course double check what I'm saying ! I would not want your work to be erased ! <3

maniospas:
Ah, sorry, my bad. Did not notice. I used GitHub's sync function, thinking that I needed to be up to date to keep compatibility with any breaking changes (if any). 😓 I'll tinker somehow. The long chain of commits above are likely those that have been merged into develop in the interim, and have no new work by me. Mine are just these two.

image

If I am unable to do anything, I will tell you; perhaps the worst case is to delete the fork and PR and create a new one, but I'd rather not resort to such ... shameful measures. XD

Becheler:
Ahhah no worries we will figure this out, this damn "Sync fork" button is a very common trap! It does a merge (not a fast-forward) when the branch has diverged, and if it gets pointed at the branch the PR is built on, it drags in all of upstream's interim history as merge noise... 😅 I don't even know why Github offers this lol

> 820c953 Merge branch 'personalized_page_rank' of github.com:maniospas/graph into develop

So it may be even worse than anticipated: if you merged your feature branch on your develop branch, it probably means your own develop (`origin`) is contaminated now, so you will need to branch of the `upstream` develop.

I suggest something around:

```
# upstream = boostorg/graph, origin = your fork
git fetch upstream

# confirm the two real commits (your work, no merges, not already on develop):
git log --no-merges --author=maniospas upstream/develop..personalized_page_rank --oneline

# fresh branch off current ****upstream**** develop:
git checkout -b pagerank-clean upstream/develop

# replay just those two SHAs, oldest first:
git cherry-pick <sha_older> <sha_newer>

# repoint the existing PR branch: keeps PR #502, no new PR needed:
git push --force-with-lease origin pagerank-clean:personalized_page_rank
```

maniospas:
Update: I think I fixed it. Sorry for spamming - thought it would be harder. Yeah, as you found out also concurrently, just cherry-picked the commits and it looks fine (thankfully, they were only a few). The new ones are `assertions` and `first directed graph tests` (the test's changes are a mess because I basically rewrote the test in grid experimentation format to try various graph and parameter combinations).

Becheler:
Amazing @maniospas ! Thanks for being so quick ! Let me have a look at those new commits ! 🙏🏽

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,248 @@
...
+
+//  Authors: Emmanouil Krasanakis
+
+#ifndef BOOST_GRAPH_PERSONALIZED_PAGE_RANK_HPP
+#define BOOST_GRAPH_PERSONALIZED_PAGE_RANK_HPP
+
+#include <boost/property_map/property_map.hpp>
+#include <boost/graph/graph_traits.hpp>
+#include <boost/graph/properties.hpp>
+#include <boost/graph/iteration_macros.hpp>
```
Becheler: dead include ?

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,248 @@
...
+#include <boost/graph/overloading.hpp>
+#include <boost/graph/detail/mpi_include.hpp>
```
Becheler: dead include ?

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,248 @@
...
+#include <boost/property_map/function_property_map.hpp>
+#include <vector>
+
```
Becheler, line 24: Several includes seem to be missing:
```cpp
#include <boost/range/iterator_range.hpp>      // boost::make_iterator_range
#include <boost/concept/assert.hpp>                // BOOST_CONCEPT_ASSERT
#include <boost/graph/graph_concepts.hpp>   // IncidenceGraphConcept, VertexListGraphConcept
#include <cmath>      // std::abs
#include <cassert>    // assert
#include <cstddef>    // std::size_t
```

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,248 @@
...
+        typename RankMap2 >
+    Done personalized_page_rank(
+        const Graph& g,
+        WeightMap weight_map,
+        PersonalizationMap personalization_map,
+        RankMap rank_map,
+        Done done,
+        typename property_traits< RankMap >::value_type damping,
+        RankMap2 rank_map2
+        BOOST_GRAPH_ENABLE_IF_MODELS_PARM(Graph, vertex_list_graph_tag))
```
Becheler: The `BOOST_GRAPH_ENABLE_IF_MODELS_PARM(Graph, vertex_list_graph_tag))` should be removed, it has no usefulness here (its for distributed stuff only, AI probably copies it by mistake from the Page Rank source code, and should never appear in public interfaces anyway). Once it's gone, the `<boost/graph/overloading.hpp>
` becomes dead too and should be gone 😺

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,248 @@
...
+        {
+            rank_type value = get(personalization_map, v)/personalization_norm;
+            put(personalization_map, v, value);
+            put(rank_map, v, value);
+        }
+        
+        bool to_map_2 = true;
+        do
+        {
+            typedef typename graph_traits< Graph >::traversal_category category;
```
Becheler: please prefer the `using category = graph_traits< Graph >::traversal_category` modern equivalent. Typedefs are things of the past 😉

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,248 @@
...
+        typename property_traits< RankMap >::value_type damping,
+        RankMap2 rank_map2
+        BOOST_GRAPH_ENABLE_IF_MODELS_PARM(Graph, vertex_list_graph_tag))
+    {
+        using Vertex = typename graph_traits<Graph>::vertex_descriptor;
+        using Edge = typename graph_traits<Graph>::edge_descriptor;
+        BOOST_CONCEPT_ASSERT(( boost::IncidenceGraphConcept<Graph> ));
+        BOOST_CONCEPT_ASSERT(( boost::VertexListGraphConcept<Graph> ));
+        BOOST_CONCEPT_ASSERT(( boost::ReadablePropertyMapConcept<WeightMap, Edge> ));
+        BOOST_CONCEPT_ASSERT(( boost::ReadablePropertyMapConcept<PersonalizationMap, Vertex>));
```
Becheler: This is underconstrained, as PerzonalizationMap uses `put` to write stuff to it to normalize in place (l. 166) and restore magnitude (l.198).
So you'd want to use `ReadWritePropertyMapConcept` that allows both for `get` and `put`

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,248 @@
...
+        BOOST_CONCEPT_ASSERT(( boost::ReadWritePropertyMapConcept<RankMap,  Vertex> ));
+
```
Becheler, line 155: But rankMap2 is completely unchecked. You will want to add:
`BOOST_CONCEPT_ASSERT(( boost::ReadWritePropertyMapConcept<RankMap2, Vertex> ));`

include/boost/graph/personalized_page_rank.hpp
```diff
@@ -0,0 +1,248 @@
...
+
+        assert (damping>=-1.0 && damping<1.0 && "Damping outside the closed-open range [-1.0,1.0) could induce numerical instability."); // non-inclussive upper limit is deliberate
```
Becheler: The convention is to rather use the Boost version of assert, that is more flexible that the standard (users can reroute it and do all sorts of tricky things):
```
#include <boost/assert.hpp> // you will need to add this
....
BOOST_ASSERT_MSG(damping >= -1.0 && damping < 1.0, "Damping outside the closed-open range [-1.0,1.0) could induce numerical instability.");
```

test/personalized_pagerank_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+        assert(convergence.get_remaining_iters()<max_iters);                                     // ran
+        assert(convergence.get_remaining_iters()>0 || (damping<0.0));                            // converged (derivatives may not converge)
+        assert((ranks[0]<ranks[1]+0.1) || (damping<0.0) || damping>1.0);                         // holds for all graphs given low-pass damping
+        assert((ranks[0]!=ranks[1]) == (damping!=0.0));                                          // but not the same normally, the same 1.0 personalization if damping is zero
+        assert(ranks[0]==ranks[2] || (damping<=-1.0));                                           // equal due to symmetry, even under non-convergence and floating coarseness
+        assert(std::abs(ranks[0]-ranks[2])<1.E-14);                                              // approximately equal in cases where even addition order matters
+        assert((ranks[0]<0.0)||(ranks[1]<0.0)||(ranks[num_vertices(g)-1]<=0.0)||(damping>0.0));  // ensure that negative flows are possible for negative damping
+    }
```
Becheler: If those tests are ever built in release mode, all those asserts will disappear 😄 
That's a reason to use Boost instead, with the following method:

```cpp
#include <boost/core/lightweight_test.hpp> // you need to include this
...
BOOST_TEST(convergence.get_remaining_iters() < max_iters); // tests it's true
...
return boost::report_errors();   // at end of main
```

Becheler: Also I guess you are aware that the entire `bidirectional_graph_tag` path is untested ? 😉

Becheler (changes requested):
All good, thanks again !

There are just more idiomatic details you sincerely had no way to be aware of. I believe I mentioned them somewhere in a contribution guideline, but there is so much to learn that you have to go through it once or twice and get bitten before to get it right 😮‍💨

include/boost/graph/personalized_page_rank.hpp
Becheler: weird display, I meant the `iteration_macros` header is dead.

include/boost/graph/personalized_page_rank.hpp
Becheler: weird display, I meant the `mpi_include` header is dead (and the overloading too, see further)

maniospas:
Thank you for the information @Becheler and going through the PR!

Submitting code with them addressed (hopefully), if you can take a look. I am aware of the bidirectional path being untested; will be writing different tests for that one to check various compliance properties between that, directional, and vanilla pagerank.

P.S. Some of those things were leftover from the original pagerank which I used as a prototype for the boilerplate (not using LLMs for things other than search in non-personal projects). For the assert, I distinctly remember reading about asserts and not being sure what to do in the end somewhere in the docs, but the contribution guidelines do contain the information, as you say, so I am not sure where my confusion came from.

Becheler:
Amazing thanks @maniospas ! 😄 
Looking at it today or tomorrow !
No worries at all, they are so many things to be aware at once that it's very complicated to do it right (even for me after several months on the project ahaha). You're doing great ! 🏆 
Yes I expected you were aware of the undirected thing - I think it's great to process step by step as you did.
Don't worry about the code coverage then, it will surely be blocking CI with a huge red cross, but just don't look at it 😉

PR #508, opened 2026-06-18
add missing const values; use some using instead of typedef

andreacassioli:
### Before submitting

- [X] I searched for an existing PR or issue covering the same change. (actually this PR superseeds https://github.com/boostorg/graph/pull/160)
- [X] My contribution is licensed under the Boost Software License 1.0.

### Type of change

- [x] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

It superseeds https://github.com/boostorg/graph/pull/160

### Motivation

Some `const` qualifiers are missing. Not a big deal, but it make it clearer to have them in place. Notice that the named parameter interface cannot be changed without introducing a breaking change: the reason is that it defaults to use graph internal property map as output, thus the graph must be modifiable to extract that map.

Incidentally I change some `typedef` to `using`.

### Testing

Existing tests should be fine

### Checklist

### DIFF
--- a/include/boost/graph/successive_shortest_path_nonnegative_weights.hpp
+++ b/include/boost/graph/successive_shortest_path_nonnegative_weights.hpp
@@ -22,6 +22,7 @@
 #include <boost/graph/properties.hpp>
 #include <boost/graph/iteration_macros.hpp>
 #include <boost/graph/detail/augment.hpp>
+#include <boost/concept/assert.hpp>
 
 namespace boost
 {
@@ -34,6 +35,13 @@ namespace detail
     : public put_get_helper< typename property_traits< Weight >::value_type,
           MapReducedWeight< Graph, Weight, Distance, Reversed > >
     {
+        
+        using edge_descriptor = typename graph_traits< Graph >::edge_descriptor;
+        using vertex_descriptor = typename graph_traits< Graph >::vertex_descriptor;
+
+        BOOST_CONCEPT_ASSERT(( ReadablePropertyMapConcept<Weight, edge_descriptor> ));
+        BOOST_CONCEPT_ASSERT(( ReadablePropertyMapConcept<Distance, vertex_descriptor> ));
+        
         using g_traits = graph_traits< Graph >;
 
     public:
@@ -55,9 +63,9 @@ namespace detail
 
     private:
         const Graph& g_;
-        const Weight weight_;
-        const Distance distance_;
-        const Reversed rev_;
+        Weight weight_;
+        Distance distance_;
+        Reversed rev_;
     };
 
     template < class Graph, class Weight, class Distance, class Reversed >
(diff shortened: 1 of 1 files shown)


include/boost/graph/successive_shortest_path_nonnegative_weights.hpp
```diff
@@ -56,9 +55,9 @@ namespace detail
 
     private:
         const Graph& g_;
-        Weight weight_;
-        Distance distance_;
-        Reversed rev_;
+        const Weight weight_;
+        const Distance distance_;
+        const Reversed rev_;
     };
```
Becheler, line 69: I don't think we should use const on private data. The argumentation around my mental model of `const` has been developed here by Arthur O'dwyer: https://quuxplusone.github.io/blog/2022/01/23/dont-const-all-the-things/

See particularly the "const members are never a good idea" section where he says constness should be enforced by the public API, not by marking private data const: https://quuxplusone.github.io/blog/2022/01/23/dont-const-all-the-things/

Specifically here, the operator is already const so it adds no security:

```cpp
reference operator[](key_type v) const {
    return get(distance_, source(v, g_)) - get(distance_, target(v, g_)) + get(weight_, v);
}
```

And more importantly, a non-static const data member deletes the implicit copy/move assignment operators, and this class is a property map created by `make_mapReducedWeight(...)` and handed to Dijkstra by value (`weight_map(...)`). Property maps are value types you copy around so making it non-assignable makes it a non-regular type for no benefit :/

Becheler (changes requested):

include/boost/graph/successive_shortest_path_nonnegative_weights.hpp
Becheler: To be clear, the other changes are great so thank you for this ! ❤️ 

In hindsight maybe what's missing is a concept check to ensure the maps are not written to ?

```cpp
BOOST_CONCEPT_ASSERT(( ReadablePropertyMapConcept<Weight, edge_descriptor> ));
BOOST_CONCEPT_ASSERT(( ReadablePropertyMapConcept<Distance, vertex_descriptor> ));
```

andreacassioli: Thanks for the pointers @Becheler I got used to maybe too many `const` lately and not much time to dig into more C++!

I have made the changes to you propose.

Becheler: No worries ! Const is tricky. I once got on a debate in industry because a guy wanted to pass parameters by const value to say "we pass it it by value and also it wont be modified in the internals". The problem being that ... it leaks internals :) 

Thank you for the changes ! I will have a last look and merge ! 🙏🏽

Becheler:
Thank you @andreacassioli for the little cleanup!
I made a last scan, the CI is green, no warnings added, the changes look good, so I will merge it 😄

PR #528, opened 2026-07-16
fix: remove dependency to MPL in graph_traits using detector idiom

Becheler:
Reducing dependencies to MPL is important for compile time, dependency graph and general hygiene, as similar behavior is available with modern C++.

### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

`graph_traits` reports a graph's associated types (vertex type, edge type...). Not every graph defines all of them, so `graph_traits` needs to check if a member type exists on the graph, use it if it does, and fall back to `void` if it doesn't.

That check was built with Boost.MPL. It used `BOOST_MPL_HAS_XXX_TRAIT_DEF` macro to generate a "has member X" trait, then `mpl::eval_if` and `mpl::identity` to pick either the member or the fallback. This pulled in five `boost/mpl` headers just to answer "does this type have member X".

This PR refactors it using standard C++ `void_t` [detector idiom](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4502.pdf). Two small template overloads do the check directly: one matches when the member type exists and exposes it, the other is the fallback. Same result, no MPL, no new dependency, behavior is unchanged.

### Motivation

Reducing dependencies to MPL is important for compile time, dependency graph and general hygiene, as similar behavior is available with modern C++.

### Checklist

### DIFF
--- a/include/boost/graph/graph_traits.hpp
+++ b/include/boost/graph/graph_traits.hpp
@@ -14,15 +14,12 @@
 #include <iterator>
 #include <utility> /* Primarily for std::pair */
 #include <boost/tuple/tuple.hpp>
-#include <boost/mpl/if.hpp>
-#include <boost/mpl/eval_if.hpp>
 #include <boost/mpl/bool.hpp>
 #include <boost/mpl/not.hpp>
-#include <boost/mpl/has_xxx.hpp>
-#include <boost/mpl/void.hpp>
-#include <boost/mpl/identity.hpp>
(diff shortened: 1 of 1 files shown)


jeremy-murphy (approved):

jeremy-murphy:
Looks good.

PR #529, opened 2026-07-16
fix: handroll mpl facilities in graphml

Becheler:
Continue removing MPL from the codebase to limit build times and dependencies.
But this time there is no direct C++ equivalent to the MPL features, so there is a small utility `graphml_detail` namespace.

### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

The GraphML reader and writer kept a compile-time list of the value types they support (`bool`, `int`, `double`, `string`, and so on) and used three Boost.MPL tools on it: 
- `mpl::vector` (the list of types), 
- `mpl::for_each` (visit each type once, to try each candidate when parsing),
-  `mpl::find` (get a type's position to look up its name). 

We added a new dependency (MP11) to use the modern equivalent.

No behavior change.

### Motivation

Part of the ongoing effort to shrink Boost.Graph's dependency footprint. Boost.MPL is an old, heavy, slow-to-compile library that modern C++ (`<type_traits>` plus small hand-written helpers) can replace directly. This removes the last direct MPL use from BGL that would not be breaking public API.

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,7 @@ target_link_libraries(boost_graph
     Boost::integer
     Boost::iterator
     Boost::lexical_cast
+    Boost::mp11
     Boost::mpl
     Boost::multi_index
     Boost::multiprecision
--- a/build.jam
+++ b/build.jam
@@ -20,6 +20,7 @@ constant boost_dependencies :
     /boost/integer//boost_integer
     /boost/iterator//boost_iterator
     /boost/lexical_cast//boost_lexical_cast
+    /boost/mp11//boost_mp11
     /boost/mpl//boost_mpl
     /boost/multi_index//boost_multi_index
     /boost/multiprecision//boost_multiprecision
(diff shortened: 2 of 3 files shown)


jeremy-murphy:
Now I realise that we're trying to reduce dependencies here, but would it be appropriate to use mp11 to replace the functionality in mpl?

Becheler:
It was originally the plan to move from MPL to MP11 because I thought the features required across BGL were non trivial.

But all changes are trivial conversions to standard equivalent. It means that pulling in MP11 would be only to avoid those 15 lines.

It would have been > 100 lines for an equivalent that may have justified pulling in MP11 but in that case I don't think it's worth, as it may as well become portable if we go C++17 (looks like the numbers of liens required in c++17 for the same behavior is divided by 2)

jeremy-murphy:
Hmmmm. I'm not really in favour of this change because it leaves a blurry line of when we would justify using mp11 for this sort of thing.
Let's make it a priority to discuss.

Becheler:
@jeremy-murphy maybe you're right. MP11 is probably pulled transitively anyway. I removed the handroll facilities and brought MP11 in.

jeremy-murphy (approved):
Ok, cool, yeah, let's prioritise getting rid of mpl, we can debate mp11 later.

PR #532, opened 2026-07-17
refactor: remove dependency to boost.math

Becheler:
### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

- remove boost.math headers and deps in CML/build.jam
- use C++11 std::hypot
- define a local pi and square root constant

### Motivation

Remove dependency on Boost.Math: it's overkill here for just one STL function and 2 constants.

### Checklist

### DIFF
--- a/include/boost/graph/circle_layout.hpp
+++ b/include/boost/graph/circle_layout.hpp
@@ -9,7 +9,6 @@
 #ifndef BOOST_GRAPH_CIRCLE_LAYOUT_HPP
 #define BOOST_GRAPH_CIRCLE_LAYOUT_HPP
 #include <boost/config/no_tr1/cmath.hpp>
-#include <boost/math/constants/constants.hpp>
 #include <boost/graph/graph_traits.hpp>
 #include <boost/graph/iteration_macros.hpp>
 #include <boost/graph/topology.hpp>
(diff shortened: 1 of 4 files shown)


jeremy-murphy (approved):
Why did you flag this as a breaking change? It doesn't seem like it to me.

Becheler:
I don't think I flagged this PR as breaking. The breaking change is in another one, where BGL public API exposed MPL types, that can now be STL types. 
Edit: oops I totally did flag it by mistake - my bad ! Friday fatigue hitting hard 😅

include/boost/graph/circle_layout.hpp
```diff
@@ -34,7 +33,7 @@ void circle_graph_layout(
 {
     BOOST_STATIC_ASSERT(
         property_traits< PositionMap >::value_type::dimensions >= 2);
-    const double pi = boost::math::constants::pi< double >();
+    const double pi = 3.14159265358979323846;
```
andreacassioli, line 36: out of curiosity: any reason for not using https://en.cppreference.com/cpp/numeric/constants ?

Becheler: They are C++20 😭😭😭
But yeah that was my hope too 😄 this std::numbers lib is dope ! Too bad it arrives so late !

andreacassioli: Of course! I should have checked before asking. Anyway, great job @Becheler !

Becheler: Thanks to you for your review !

PR #533, opened 2026-07-17
refactor: drop direct dependency on Boost.Spirit (graphviz parser)

Becheler:
### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

- `graphviz.hpp`: removed both Boost.Spirit code paths in the two `read_graphviz` overloads (they now always call `read_graphviz_new`)
- dropped the unconditional `classic_multi_pass` Spirit include.
- `BOOST_GRAPH_USE_SPIRIT_PARSER` still recognized, but now emits a `BOOST_PRAGMA_MESSAGE` deprecation notice and redirects to the default parser instead of selecting Spirit.
- Deleted `detail/read_graphviz_spirit.hpp` (the Spirit DOT reader that pulled in 10 `boost/spirit` headers).
- `test/isomorphism.cpp`: removed the incidental `#define BOOST_GRAPH_USE_SPIRIT_PARSER`
- `build.jam` and `CMakeLists.txt` dropped the `spirit` dependency.

### Motivation

The spirit-based graphviz parser has been non default, undocumented and untested for 16 years. 
The Boost.Spirit library has 28 direct boost dependencies, and 54 indirect, effectively dragging most of Boost in BGL: https://alandefreitas.github.io/boostdep_graph/libs/spirit.html

This PR removes its implementation, effectively dropping spirit dependency and its direct+transitive chain.

The default parser uses Boost.Regex, that has a much better dependency chain: https://alandefreitas.github.io/boostdep_graph/libs/regex.html

Note: spirit is not effectively removed from the transitive dependencies chain, as Boost.Serialization pulls it in.
A next PR will drop Boost.Serialization for Boost.Core.Serialization.

Note: Boost.Spirit has several dependants that Boost.Graph uses:
- Boost.PropertyTree
- Boost.MPI

### Checklist


mborland (approved):
LGTM. Very straightforward, and helps remove a heavy dependency.

include/boost/graph/graphviz.hpp
```diff
@@ -925,13 +925,12 @@ namespace detail
 } // end namespace boost::detail::graph
 
 #ifdef BOOST_GRAPH_USE_SPIRIT_PARSER
-#ifndef BOOST_GRAPH_READ_GRAPHVIZ_ITERATORS
-#define BOOST_GRAPH_READ_GRAPHVIZ_ITERATORS
+BOOST_PRAGMA_MESSAGE(
+    "BOOST_GRAPH_USE_SPIRIT_PARSER is deprecated and no longer has any effect: "
```
jeremy-murphy, line 929: "Deprecated" means that something still exists (and works) but its use is not recommended, so it's not really the right meaning here. I think that text can simply be removed.
```suggestion
    "BOOST_GRAPH_USE_SPIRIT_PARSER no longer has any effect: "
```

jeremy-murphy (commented):
One pedantic post-merge request.  :)

include/boost/graph/graphviz.hpp
Becheler: You're right! Fixed in #584 
I added a comment with `deprecated` and `1.95` because I will use this in a grep tag to track what must be removed in 1.95.

jeremy-murphy:
> Boost dependency footprint vs `develop` (auto-generated). PR run `33410598370` vs develop run `33397888958` (`92470302c3`).
> 
> **Header-inclusion weights** (graph files pulling each direct dependency in):
> 
> Dependency	develop	PR	Δ
> spirit	2	0	-2
> config	94	93	-1
> core	15	14	-1
> function	2	1	-1
> property_map	80	79	-1
> throw_exception	11	10	-1
> type_traits	42	41	-1
> **Transitive Boost modules:** 66 → 66 (0)

So, removing Boost.Spirit didn't actually remove our dependency on any transitive modules?

Becheler:
Correct, and it was expected (see the Notes in the PR description), because as long as Boost.PropertyTree is still a direct dependency, it will pull both Boost.Spirit and Boost.Serialization into the transitive dependency chain (those two libs are 55/68 = 80% of the dependency chain weight)

https://github.com/boostorg/graph/pull/547 Aims at getting rid of that, with additional benefits:

- x3.5 faster runtime
- x3.7 less memory
- faster compilation time 

I am running benchmarks on project compilation times before/after dependencies removal as we speak (compile time dominated by transitive header count). The figure is still a bit shitty, but interesting and kinda confirms that graphviz pulls half the boost galaxy and motivated #566 :

image

jeremy-murphy:
> https://github.com/boostorg/graph/pull/547 Aims at getting rid of that, with additional benefits:
> 
> x3.5 faster runtime
> x3.7 less memory
> faster compilation time

For what part of the library exactly? Just for GraphML or more broadly?

PR #535, opened 2026-07-20
drop dependency on Boost.Bimap and Boost.Lambda

Becheler:
Drop Boost.Bimap dependency

### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

Drop the Boost.Bimap dependency, adds a thin wrapper with similar interface.

### Motivation

`two_graphs_common_spanning_trees.hpp` is the only user of Bimap in Boost.Graph (and actually in the entire Boost ecosystem).

- **It pulls 26 modules** [(`mpl`, `multi_index`, `mp11`, `lambda`, ...)](https://alandefreitas.github.io/boostdep_graph/libs/bimap.html) for a container used read-only. It's a blocker to get rid of MPL and other deprecated Boost libs since they will keep popping up through the transitive dependencies chain.
- **Most of the interface is unused.** Each bimap is built once then read via four ops: `left.at`, `right.at`, `right.find`, `insert`.
- **Bimap is over-general.** The position side is a dense range `0..n-1` (plain array indexing), so only edge-to-position needs a map. A `std::vector` plus a `std::map` cover both directions; a bidirectional-map library is the wrong tool.

### Testing

Same tests hold.

### Checklist

### DIFF
--- a/include/boost/graph/two_graphs_common_spanning_trees.hpp
+++ b/include/boost/graph/two_graphs_common_spanning_trees.hpp
@@ -13,7 +13,6 @@
 
 #include <boost/config.hpp>
 
-#include <boost/bimap.hpp>
 #include <boost/type_traits.hpp>
 #include <boost/concept/requires.hpp>
 #include <boost/graph/graph_traits.hpp>
@@ -23,6 +22,8 @@
 #include <vector>
 #include <stack>
 #include <map>
+#include <utility>
+#include <cstddef>
 
 namespace boost
 {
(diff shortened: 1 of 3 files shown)


include/boost/graph/two_graphs_common_spanning_trees.hpp
```diff
@@ -123,14 +124,43 @@ namespace detail
...
+    // Position/edge lookup used by the algorithm below. left maps a position
+    // index to its edge, right maps an edge back to its position.
+    template < typename Edge > struct index_bimap
+    {
+        using value_type = std::pair< std::size_t, Edge >;
+
+        struct left_type
+        {
+            std::vector< Edge > store;
+            const Edge& at(std::size_t i) const { return store.at(i); }
```
jeremy-murphy: Why are you using `at`?

Becheler: Would you prefer this (since the indices are guaranteed to be in [0..n-1]) ?
```
const Edge& at(std::size_t i) const { return store[i]; }
```

jeremy-murphy: Generally, yes. `at` is for when you're not really sure how big your container is.

jeremy-murphy (commented):
I just want to look at this on a proper computer screen but I'm feeling very favourable towards it. 😅

Becheler:
Take your time the PR is not going anywhere ahaha

include/boost/graph/two_graphs_common_spanning_trees.hpp
```diff
@@ -123,14 +124,44 @@ namespace detail
...
+    // index to its edge, right maps an edge back to its position.
+    template < typename Edge > struct index_bimap
+    {
+        using value_type = std::pair< std::size_t, Edge >;
+
+        struct left_type
+        {
+            std::vector< Edge > store;
+            // i is a dense position in [0, n), so unchecked access is safe.
+            const Edge& at(std::size_t i) const { return store[i]; }
```
jeremy-murphy: By convention, `at(std::size_t i)` is checked access, meaning it will throw an exception if `i` is out of range. Ideally, we want to provide `operator[](std::size_t i)`, which is unchecked, although I can see that will mean a lot of trivial changes from `at(i)` to `[i]`.

Becheler: I think I disagree with this direction, that is modifying the public API semantics based on a change of the implementation detail. 

The point of the wrapper was to serve as a switch with the old Bimap interface. I do think we should still respect its semantics (who knows, if std::bimap arrives someday).

Concretely, for a map the semantics is different between `at()` and `operator[]`, particularly concerning insertion/throwing.

I was ok changing the internal access because we have the guarantee it will remain in scope, but from the calling site perspective we must preserve the semantics that we don't intend to insert an element (or whatever UB Bimap would be doing if unchecked access) if it does not exist.

Happy to discuss though, as you said its a trivial change, but I keep thinking they had a reason to use the `at(i)` rather than `[i]`

Becheler: @jeremy-murphy I'm not too sure who is right or wrong anymore. I updated the code to account for your remarks 😄

PR #536, opened 2026-07-20
[hawick] Support passing std::reference_wrapper to boost::hawick_circuits

ldionne:
Reported as ldionne/hawick_circuits#1.

### DIFF
--- a/include/boost/graph/hawick_circuits.hpp
+++ b/include/boost/graph/hawick_circuits.hpp
@@ -94,12 +94,10 @@ namespace hawick_circuits_detail
         typedef T type;
     };
 
-#if __cplusplus >= 201103L
     template < typename T >
     struct unwrap_reference_wrapper<std::reference_wrapper<T> > {
         typedef T& type;
     };
-#endif
 
     /*!
      * @internal
--- a/include/boost/graph/hawick_circuits.hpp
+++ b/include/boost/graph/hawick_circuits.hpp
@@ -20,6 +20,7 @@
 #include <boost/tuple/tuple.hpp> // for boost::tie
 #include <boost/type_traits/remove_reference.hpp>
 #include <boost/utility/result_of.hpp>
+#include <functional> // for std::reference_wrapper
 #include <set>
 #include <utility> // for std::pair
 #include <vector>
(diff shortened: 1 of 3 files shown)


ldionne:
This supersedes #249 which I can't modify anymore for some reason.

ldionne:
@Becheler I created this PR from my own fork, so you can clean up any branches on the boostorg repo if desired.

include/boost/graph/hawick_circuits.hpp
```diff
@@ -88,6 +89,18 @@ namespace hawick_circuits_detail
         return std::find(boost::begin(c), boost::end(c), v) != boost::end(c);
     }
 
+    template < typename T >
+    struct unwrap_reference_wrapper {
+        typedef T type;
+    };
+
+#if __cplusplus >= 201103L
```
Becheler: BGL bumped to C++14 so this guard looks like dead code now ?

test/hawick_circuits.cpp
```diff
@@ -62,6 +78,26 @@ int main()
         cycle_test(call_hawick_unique_circuits(ml), nc3[ml], nc4[ml]);
     }
 
+    // Make sure we can pass a reference_wrapper to the algorithm.
+#if __cplusplus >= 201103L
```
Becheler: same remark, this is probably obsolete now with a C++14 baseline ?

Becheler (changes requested):
Other than just an obsolete guard, I think this is good to go ? 
Thank you so much for your reactivity Louis ! 🙏🏽 
Good to finally interact after seing your name everywhere ahahaha ❤️

jeremy-murphy:
> Other than just an obsolete guard, I think this is good to go ? 
> Thank you so much for your reactivity Louis ! 🙏🏽 
> Good to finally interact after seing your name everywhere ahahaha ❤️ 

Thanks for refreshing your PR, @ldionne. The only other modernization I would make right now is to use `= delete` on that copy constructor instead of making it private. Everything else I think we can do in a batch one day.

ldionne:
Comments applied, thanks for the review and for the kind words!

jeremy-murphy (approved):

Becheler (approved):

PR #537, opened 2026-07-21
refactor: drop direct dependency on Boost.SmartPtr

Becheler:
General hygiene related to Boost.SmartPtr

### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

- Remove dead includes
- Refactor smart pointers from Boost to STL
- Removed the dead `r_c_shortest_paths_label` base `boost::enable_shared_from_this`. `shared_from_this()` was never called.
- Deleted 4 free comparison operators for `shared_ptr<label>` that dereferenced the pointer to sort (by label). With `std::shared_ptr` those would compete with STL comparison (by pointer).
- Replaced them with comparators dereferencing the pointer: a small `splabel_greater` functor for the `priority_queue`, and a lambda for `list::sort`.

### Motivation

Boost.SmartPtr does not have heavy dependencies (config, assert, throw_excpetion and core), but dead includes and C++14 STL is proper hygiene: https://github.com/cppalliance/cppalliance.github.io

### Testing

Unchanged tests in CI

### Checklist

### DIFF
--- a/include/boost/graph/r_c_shortest_paths.hpp
+++ b/include/boost/graph/r_c_shortest_paths.hpp
@@ -13,8 +13,7 @@
 #include <vector>
 #include <list>
 
-#include <boost/make_shared.hpp>
-#include <boost/enable_shared_from_this.hpp>
+#include <memory>
 #include <boost/graph/graph_traits.hpp>
 #include <boost/graph/iteration_macros.hpp>
 #include <boost/property_map/property_map.hpp>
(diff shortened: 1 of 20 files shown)


include/boost/graph/r_c_shortest_paths.hpp
```diff
@@ -119,49 +116,18 @@ inline bool operator>=(
...
-        r_c_shortest_paths_label< Graph, Resource_Container > >& u)
-{
-    return *t >= *u;
-}
-
 namespace detail
 {
 
+    // Order Splabels by the pointed-to label, not by pointer identity.
+    template < class Splabel > struct splabel_greater
```
jeremy-murphy: Use a more general name than `Splabel` as this actually works on anything that can be dereferenced.
`Pointer` is fine by me but maybe there is something better.

jeremy-murphy (approved):
Very minor request, so I'm approving now.

include/boost/graph/r_c_shortest_paths.hpp
Becheler: I went with:
```cpp
   template < class Pointer > struct deref_greater
    {
        bool operator()(const Pointer& a, const Pointer& b) const
        {
            return *a > *b;
        }
    };
```

PR #538, opened 2026-07-23
Rcsp new interface overloads

andreacassioli:
### Before submitting

### Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

- Introduce a new set of interfaces that drop the unused edge index map. The old interfaces are kept but re-wired to the new ones.
- tests are duplicated introducing a dedicated file for the old interface
- examples are duplicated introducing a dedicated file for the old interface
- No logic changes.

### Motivation

The current interfaces expose a redundant parameter (edge index map)

- interface cluttering
- less readable code
- less clear behavior (why is that parameter needed?)

The issue has been raised here https://github.com/boostorg/graph/issues/342

### Checklist

- [X] Existing tests pass (`b2` in the `test/` directory).


Becheler:
Hi @andreacassioli !

Looks good to me, great work! 

To tie it up, we would want to add deprecation notices on the old interfaces as mentioned in the previous conversation so users get a warning (note their position between template args and return type):

```cpp
#include <boost/config.hpp>

template < class Graph, class VertexIndexMap, class EdgeIndexMap, class Resource_Container, class  Resource_Extension_Function, class Dominance_Function >
BOOST_DEPRECATED("edge_index_map is unused; drop it and call the 5-argument r_c_shortest_paths overload. Removal planned for Boost 1.95.")
void r_c_shortest_paths(const Graph& g, const VertexIndexMap& vertex_index_map, const EdgeIndexMap& edge_index_map, ...)
{
    r_c_shortest_paths(g, vertex_index_map, s, t, ...);
}
```

Old examples and test files should begin with a `#define BOOST_ALLOW_DEPRECATED_SYMBOLS` on their first line (before any include) to disable warnings (we know they are outdated and should soon be removed).

See demonstration here: https://godbolt.org/z/dYszsMaea

I think it's reasonable to duplicate the unit tests and the example here (at the cost of blowing up the PR size but there it's a transient inconvenient). If you could quickly confirm locally that the old and new example generate the same output that would be gold! 🙏🏽 

Let me know if you have any question,
Thanks again for this cleanup help !
Arno

Becheler:
I had a look at why the warning count increased, it confirmed my suspicion it was not because of this PR but rather prexisting warnings on develop that got duplicated by duplicating `.cpp` files. 

Considering we want to separate deprecation PRs from cleanup PRs, I suggest those can be fixed in a different PR.

### Dangling reference

```
test/r_c_shortest_paths_test.cpp:160:45: warning: possibly dangling reference to a temporary [-Wdangling-reference]
  160 |         const SPPRC_Example_Graph_Arc_Prop& arc_prop = get(edge_bundle, g)[ed];
```
Proposed fix:  replace `const SPPRC_Example_Graph_Arc_Prop& arc_prop = get(edge_bundle, g)[ed];` by `const SPPRC_Example_Graph_Arc_Prop arc_prop = get(edge_bundle, g)[ed];` since property maps are by value anyway (two occurences line 160 and 264 in the test file, and 1 occurence in the example file).

### Unused type alias

```
../../../boost/graph/r_c_shortest_paths.hpp:166:49: warning: typedef ‘LTraits’ locally defined but not used [-Wunused-local-typedefs]
```

Proposed fix: delete the dead alias.

andreacassioli:
@Becheler  I have incorporated the suggestions. Also I have modified the docs for just one interface. Not sure how it should look like in this case. Could you take a look?

doc/modules/ROOT/examples/algorithms/layout/kamada_kawai.txt
```diff
@@ -1,6 +1,6 @@
 converged: yes
```
Becheler, line 1: this looks like examples were regenerated and committed by mistake ? Can you remove those two `.txt` changes ? Otherwise it looks great ! 😄

Becheler (changes requested):

doc/modules/ROOT/examples/algorithms/layout/kamada_kawai.txt
andreacassioli: done

andreacassioli:
@Becheler I have updated the docs, but one table refuses to align and overlaps with the Contents. I have tried the autowrap parameter, but no way. Suggestions?

Becheler:
Thank you Andrea,

I would say don't worry about it and make the minimal formatting mimicking what was originally in the documentation.

It is not clear yet what the documentation infrastructure pulls from the bundle, or if local builds differ from online rendering and in what ways. 

If it is very problematic I will ship a quick PR to format it after we merge yours :)

andreacassioli:
> Thank you Andrea,
> 
> I would say don't worry about it and make the minimal formatting mimicking what was originally in the documentation.
> 
> It is not clear yet what the documentation infrastructure pulls from the bundle, or if local builds differ from online rendering and in what ways.
> 
> If it is very problematic I will ship a quick PR to format it after we merge yours :)

The issue is one table of function parameter is overflowing a bit over the context bar on the right. Not the big deal I would say.

Becheler:
Ok are you ok to merge it as is ? I will see what i can do to fix the overlap tomorrow 😄  
PS: also you can add your name in the copyright since you modified it (if you want) 😉

andreacassioli:
I am OK to merge.

Becheler:
Amazing ! Thank you @andreacassioli ! 🥳

PR #547, opened 2026-08-05
refactor: drop property_tree dependency

Becheler:
### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [x] Other (specify below)

- [x] Yes (describe migration impact below)
- [x] No

### What this PR does

- Remove dependency to Boost.PropertyTree and its [58 transitive dependencies](https://alandefreitas.github.io/boostdep_graph/libs/property_tree.html)
- Vendor the core BGL uses (rapidxml header) with no behavioral change
- Performance on local benchmark against develop:
    - x3.5 faster
    - x3.7 less memory
    
 > @jeremy-murphy : "It's not a refactor because of the change to the exception type. That's not a blocker, it just means we need to treat this change appropriately such as in the release notes make it clear about what the breaking change is."
 
### Motivation

- GraphML was originally written in 2007 by Tiago de Paula Peixoto using expat, an external C XML library with a handmanaged SAX callback parser. 
- In 2009, Jeremiah Willcock rewrote it onto Boost.PropertyTree (commit e5aebbcc, "fixes #3300"), which removed the external expat dependency and cut the parser from ~357 lines to ~97 by replacing the SAX state machine with a simple tree walk. 
- But in 2026 property_tree becomes the heavy dependency worth dropping.
- Rapidxml.hpp is the only real dependency. There is no need to bring property_tree itself, on the contrary: it copies the whole tree again and is largely inefficient.

property_tree's value is a uniform generic tree with a string-path query API, the same interface we would use for JSON or INI. But GraphML never needs that generality: it only ever asks "tag name? this attribute? this text?", which a rapidxml node answers directly.

Under the hood, property_tree already parses XML with rapidxml (nodes are non owning, pointers to buffer), then copies the entire parse result into its generic ptree so the uniform API works (pure overhead with many allocations as nodes are recursive pairs `string`/`ptree`).

This PR keeps the exact same rapidxml parser but walks its DOM directly, deleting the ptree copy (and with it the property_tree dependency).

### Testing

The legacy tests properly exercise the round trip and code path to guarantee no regression. The only legacy behavior we can't match is the exception type being raised. Malformed XML now throws `boost::parse_error` instead of property_tree's `xml_parser_error`/`ptree_bad_path`

Benchmark script to compile and run on develop then on the PR branch:

```cpp
#include <boost/graph/graphml.hpp>
#include <boost/graph/random.hpp>
#include <boost/random/mersenne_twister.hpp>

#include <algorithm>
#include <chrono>
#include <iostream>
#include <sstream>
#include <string>
#include <vector>

#include <sys/resource.h>

constexpr std::size_t NVERTICES = 100000;
constexpr std::size_t NEDGES = 1000000;
constexpr int REPS = 5;

using graph_t = boost::adjacency_list< boost::vecS, boost::vecS,
    boost::directedS,
    boost::property< boost::vertex_name_t, std::string,
        boost::property< boost::vertex_color_t, int > >,
    boost::property< boost::edge_weight_t, double > >;

static std::string make_graphml()
{
    graph_t g;
    boost::mt19937 rng(42);
    boost::generate_random_graph(g, NVERTICES, NEDGES, rng, false, false);

    std::size_t i = 0;
    graph_t::vertex_iterator vi, ve;
    for (boost::tie(vi, ve) = vertices(g); vi != ve; ++vi, ++i)
    {
        boost::put(boost::vertex_name_t(), g, *vi, "node <" + std::to_string(i) + "> & \"x\"");
        boost::put(boost::vertex_color_t(), g, *vi, int(i % 255));
    }
    double w = 0.5;
    graph_t::edge_iterator ei, ee;
    for (boost::tie(ei, ee) = edges(g); ei != ee; ++ei)
        boost::put(boost::edge_weight_t(), g, *ei, w += 0.001);

    boost::dynamic_properties dp;
    dp.property("name", boost::get(boost::vertex_name_t(), g));
    dp.property("color", boost::get(boost::vertex_color_t(), g));
    dp.property("weight", boost::get(boost::edge_weight_t(), g));
    std::ostringstream out;
    boost::write_graphml(out, g, dp, true);
    return out.str();
}

int main()
{
    std::string xml = make_graphml();
    const double mb = xml.size() / (1024.0 * 1024.0);

    std::vector< double > ms;
    for (int r = 0; r < REPS; ++r)
    {
        graph_t g;
        boost::dynamic_properties dp(boost::ignore_other_properties);
        dp.property("name", boost::get(boost::vertex_name_t(), g));
        dp.property("color", boost::get(boost::vertex_color_t(), g));
        dp.property("weight", boost::get(boost::edge_weight_t(), g));
        std::istringstream in(xml);
        auto t0 = std::chrono::steady_clock::now();
        boost::read_graphml(in, g, dp);
        auto t1 = std::chrono::steady_clock::now();
        ms.push_back( td::chrono::duration< double, std::milli >(t1 - t0).count());
    }
    std::sort(ms.begin(), ms.end());

    rusage ru{};
    getrusage(RUSAGE_SELF, &ru);
#ifdef __APPLE__
    double peak_mb = ru.ru_maxrss / (1024.0 * 1024.0);
#else
    double peak_mb = ru.ru_maxrss / 1024.0;
#endif

    std::cout << "size=" << mb << "MB  read median=" << ms[ms.size() / 2]  << "ms  peakRSS=" << peak_mb << "MB\n";
    return 0;
}
```
On develop:
```
size=113.66MB  read median=6382.51ms  peakRSS=5388.23MB
```
On PR:
```
size=113.66MB  read median=1825.4ms  peakRSS=1463.05MB
```

### Checklist

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -33,3 +33,5 @@ comment:
 ignore:
   - example/**
   - test/**
+  # vendored third-party rapidxml parser, not our code to test
+  - include/boost/graph/detail/rapidxml.hpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,6 @@ target_link_libraries(boost_graph
     Boost::parameter
     Boost::preprocessor
     Boost::property_map
-    Boost::property_tree
     Boost::random
     Boost::range
     Boost::serialization
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,6 @@ target_link_libraries(boost_graph
     Boost::parameter
     Boost::preprocessor
     Boost::property_map
-    Boost::property_tree
     Boost::random
     Boost::range
     Boost::serialization
--- a/build.jam
+++ b/build.jam
@@ -28,7 +28,6 @@ constant boost_dependencies :
     /boost/parameter//boost_parameter
     /boost/preprocessor//boost_preprocessor
     /boost/property_map//boost_property_map
-    /boost/property_tree//boost_property_tree
     /boost/random//boost_random
     /boost/range//boost_range
     /boost/serialization//boost_serialization
--- a/build.jam
+++ b/build.jam
@@ -27,7 +27,6 @@ constant boost_dependencies :
     /boost/parameter//boost_parameter
     /boost/preprocessor//boost_preprocessor
     /boost/property_map//boost_property_map
-    /boost/property_tree//boost_property_tree
     /boost/random//boost_random
     /boost/range//boost_range
     /boost/serialization//boost_serialization
--- a/include/boost/graph/detail/rapidxml.hpp
+++ b/include/boost/graph/detail/rapidxml.hpp
@@ -1,0 +1,2606 @@
+// ----------------------------------------------------------------------------
+// Copyright (C) 2006, 2009 Marcin Kalicinski
+//
+// Distributed under the Boost Software License, Version 1.0. 
+// (See accompanying file LICENSE_1_0.txt or copy at 
+// http://www.boost.org/LICENSE_1_0.txt)
+//
+// For more information, see www.boost.org
+// ----------------------------------------------------------------------------
+//
+// This is a vendored copy of the rapidxml parser that Boost.PropertyTree
+// carries at boost/property_tree/detail/rapidxml.hpp. It lets Boost.Graph's
+// GraphML reader parse XML without depending on the property_tree module.
+// The only edits from the upstream copy are mechanical, to avoid clashing
+// with a translation unit that also includes property_tree: the namespace is
+// boost::graph::detail::rapidxml, the include guard and the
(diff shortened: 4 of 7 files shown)


jeremy-murphy:
It's not a refactor because of the change to the exception type. That's not a blocker, it just means we need to treat this change appropriately such as in the release notes make it clear about what the breaking change is.

jeremy-murphy:
Anyway, looks like you've got your work cut out to get the coverage back up.  :)

Becheler:
@jeremy-murphy thank you for the feedback !

Code coverage was actually never 100% so any PR touching uncovered code must cover it, it's cumbersome albeit necessary ahah 😆 Coverage is now up, with a single missing line partially covered, I suggest to leave it as is.

1. Do you think this specific API breaking requires a deprecation cycle or release note + a warning in the documentation will be enough ? 
2. Also what is your favored way to accumulate release notes during in between releases ? A pinned discussion on the Github discussion maybe ?

PR #569, opened 2026-08-20
test: filling gaps in MAS and SW

Becheler:
Adding test support before tackling fixes:
#297 
#286 

### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [x] Other (specify below)

Adding tests.

- [x] No

### Checklist

### DIFF
--- a/test/stoer_wagner_test.cpp
+++ b/test/stoer_wagner_test.cpp
@@ -23,12 +23,11 @@
 #include <boost/core/lightweight_test.hpp>
 #include <boost/tuple/tuple.hpp>
 
-typedef boost::adjacency_list< boost::vecS, boost::vecS, boost::undirectedS,
-    boost::no_property, boost::property< boost::edge_weight_t, int > >
-    undirected_graph;
-typedef boost::property_map< undirected_graph, boost::edge_weight_t >::type
-    weight_map_type;
-typedef boost::property_traits< weight_map_type >::value_type weight_type;
+#include "mas_sw_maxflow_oracle.hpp"
+
+using mas_sw_oracle::undirected_graph;
+using mas_sw_oracle::weight_map_type;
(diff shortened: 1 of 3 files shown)


test/stoer_wagner_test.cpp
```diff
@@ -41,6 +43,32 @@ struct edge_t
...
+{
+    const std::size_t sizes[] = { 6, 8, 10, 12 };
+    for (std::size_t si = 0; si < sizeof(sizes) / sizeof(sizes[0]); ++si)
+    {
+        const std::size_t n = sizes[si];
+        for (std::size_t seed = 1; seed <= 25; ++seed)
+        {
+            std::mt19937 rng(static_cast< std::mt19937::result_type >(seed));
+            const undirected_graph g = mas_sw_oracle::make_random_connected_graph(n, n, rng);
+            BOOST_TEST(mas_sw_oracle::is_connected(g));
```
jeremy-murphy: I know this practice is probably widespread in BGL, but I'm generally against using randomness in unit tests. What we want is careful selection of inputs that trigger different conditions inside the algorithms. Fuzzing has its place, but not here.

jeremy-murphy (changes requested):

test/stoer_wagner_test.cpp
Becheler: Oh ? You want to manually define inputs ? I think I agree with the idea, but practically what if outputs are not trivial given the input ? And how to test for scale effects ?

Becheler: I worked on small graphs, deterministic, and will merge the issue because I need it for further work. My udnerstanding is that I answered your expectation but tell me if not and I will fix it 😄

PR #585, opened 2026-09-02
doc: reinstate stanford graph base documentation page with example

Becheler:
### Before submitting

### Type of change

- [ ] Bug fix
- [x] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

Reinstate the lost documentation page: https://www.boost.org/doc/libs/1_34_0/libs/graph/doc/stanford_graph.html

### Checklist


Becheler:
@ascherer the SGB interface documentation is back and available here: https://585.graph.prtest3.cppalliance.org/libs/graph/doc/html/graph/extending/stanford_graph.html

Is there anything you would like to see ? 😄

ascherer:
@Becheler Thanks for the restauration. I have two annotations:

1. My updated SGB creates a dynamic shared object `libgb.so` (located at `/usr/lib/x86_64-linux-gnu/sgb/libgb.so` when installed from `sgb_2\:20251228-51_amd64.deb` on Kubuntu). Check **Using the SGB** and **Where Defined**.
2. Paragraph **Missing include guards** can be deleted; the modernized SGB has custom include guards for all `gb_*` modules, e.g., `GB_IO_H` in `sgb/gb_io.h`.

I am contemplating to suggest a major change in `boost/graph/stanford_graph.hpp`: The updated `Makefile` of SGB places the header files in `/usr/include/sgb`, so the `#include` statements in the BGL header could be extended with `sgb/` for every `gb_*` interface. This would spare users from having to give the `-I /usr/include/sgb` option to the C++ ompiler. (I don't know how to resolve `-L /usr/lib/x86_64-linux-gnu/sgb` for `-lgb` automatically. It seems to be a `debbuild` thing to add the machine type.) WDYT?

Becheler:
Thank you @ascherer 

In the meantime I modernized one minimal example to run in our runnable CI examples `doc/module/examples`. Examples in BGL have never been exercised by CI and so many have broken along the years. I am not sure we want BGL unit tests to depend on an external dependency, so I left it for documentation CI job only.

I also added a small section describing to users what this SGB page is (not) about.

https://585.graph.prtest3.cppalliance.org/libs/graph/doc/html/graph/extending/stanford_graph.html

About your changes, I would be careful. 

I agree that having `#include <sgb/gb_graph.h>` instead of `#include <gb_graph.h>` would be cosmetically nicer.

However changing the headers of BGL would break users including our CI since we rely on `apt-get` package. I could use the new package, but it's more fragile and harder to maintain. I could add a symbolic link in CI to fake the new layout, but that will still break users. If there are now two layouts in the wild, then maybe BGL needs to adapt to both:

```cpp
#if defined(__has_include) && __has_include(<sgb/gb_graph.h>)
#  // use the sgb/ prefix
#else
#  // use bare names
#endif
```

Maybe we could add something to it like `#warning "the unprefixed SGB header layout is deprecated. Install an SGB that uses <sgb/...>"` but it would mean BGL would be deprecating other people's packaging conventions from a downstream header... I don't think it's the right path. Plus the old layout will still remain in Debian unless their maintainers decide differently, right ?

So in conclusion I do think the compatibility shim is the best of two worlds.

ascherer:
> However changing the headers of BGL would break users

Unfortunately, yes. Let's keep the interface intact.

Becheler:
> Unfortunately, yes. Let's keep the interface intact.

@ascherer  Sorry if I was not clear, but unless I'm mistaken, this would support both layouts and not break anyone ?
So users would be free to use both packages/layouts and the BGL header adapts ?

```cpp
#if defined(__has_include) && __has_include(<sgb/gb_graph.h>)
  // use the sgb/ prefix
  #include <sgb/gb_graph.h>
#else
  // use bare names
  #include <gb_graph.h>
#endif
```

ascherer:
@Becheler I'm not familiar with `__has_include`, but GNU has this caveat: https://gcc.gnu.org/onlinedocs/gcc-15.3.0/cpp/_005f_005fhas_005finclude.html

Becheler:
Good catch. Thinking about it, BGL is now C++14, and supported compiler/platforms (at least those in CI) all support the symbol IIUC.

After looking around, it's also used across Boost Asio and Config:

https://github.com/boostorg/asio/blob/a7dc25b4cb6c49a6946d86ea20664f1027203225/include/boost/asio/detail/config.hpp#L491-L503

https://github.com/boostorg/config/blob/93861f26d575b2952e1af8edd5bf84a1d82163c2/include/boost/config/detail/suffix.hpp#L1156-L1173

ascherer:
@Becheler As I said, I'm not familiar with `__has_include`, but **cppreference.com** lists it as standard from **C++17** and as a _comforming extension_ in **C++14**: https://en.cppreference.com/cpp/preprocessor/include

Anyway, feel free to add the appropriate preprocessor logic in `stanford_graph.hpp` as you see fit.

PR #590, opened 2026-09-06
refactor: drop Boost.Algorithm

Becheler:
### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

- `topology.hpp`: replaced `algorithm/minmax.hpp` include with `<algorithm>`
- `howard_cycle_ratio.hpp` and `bc_clustering.hpp`: replaced `boost::first_max_element` with an explicit first-max loop
- `graphviz.hpp`: replaced `boost::algorithm::replace_all` with a quote-escape loop
- `read_graphviz_new.cpp`: replaced `boost::algorithm::to_lower_copy` with `std::transform`
- `parallel_edges_loops_test.cpp` and `all_planar_input_files_test.cpp`: replaced `boost::split` with a loop
- Added `test/bc_clustering_test.cpp`: guards the max-edge pick (bridge removed, graph splits into 2 components)

`std::max_element` is not usable here: it requires a ForwardIterator, and libc++ enforces that with a hard static_assert, but BGL's edge iterators declare a category that does not satisfy it. `boost::first_max_element` had no such check, so the replacement is a short explicit loop instead.

### Motivation

Boost.Algorithm adds 30 dependencies for something that STL and short loops can do easily.

### Checklist

### DIFF
--- a/include/boost/graph/bc_clustering.hpp
+++ b/include/boost/graph/bc_clustering.hpp
@@ -9,7 +9,6 @@
 #ifndef BOOST_GRAPH_BETWEENNESS_CENTRALITY_CLUSTERING_HPP
 #define BOOST_GRAPH_BETWEENNESS_CENTRALITY_CLUSTERING_HPP
 
-#include <boost/algorithm/minmax_element.hpp>
 #include <boost/graph/betweenness_centrality.hpp>
 #include <boost/graph/graph_traits.hpp>
 #include <boost/graph/graph_utility.hpp>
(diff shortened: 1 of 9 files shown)


include/boost/graph/bc_clustering.hpp
```diff
@@ -132,8 +131,11 @@ void betweenness_centrality_clustering(MutableGraph& g, Done done,
             edge_centrality_map(edge_centrality)
                 .vertex_index_map(vertex_index));
         std::pair< edge_iterator, edge_iterator > edges_iters = edges(g);
-        edge_descriptor e
-            = *boost::first_max_element(edges_iters.first, edges_iters.second, cmp);
+        auto max_edge_it = edges_iters.first;
+        for (auto edge_it = edges_iters.first; edge_it != edges_iters.second; ++edge_it)
+            if (cmp(*max_edge_it, *edge_it))
+                max_edge_it = edge_it;
+        edge_descriptor e = *max_edge_it;
```
jeremy-murphy, line 138: std::max_element - cppreference.com https://share.google/cBYeq3gW04DAipAiH
?

jeremy-murphy (changes requested):

include/boost/graph/bc_clustering.hpp
Becheler: That was my first try but it fails compilation because of a conflict between requirements of boost versus stl iterators (ForwardIterator).

Becheler: Passing to std would be going back to the problem described in #175 and #190

jeremy-murphy: Ehhh, I see, thanks.

Issue #593, opened 2026-09-07
[draft] Good first issues: start contributing to Boost.Graph here


Becheler:
Welcome! 🥳 

This is the landing issue for anyone who wants to make a first contribution to Boost.Graph. Every task below is small, self-contained, and reviewable in one sitting. They are real cleanups we want done, not busywork.
 
Tasks are grouped into three tiers by the C++ background they actually require, so you can pick one that matches where you are. 
- Tier 1 needs nothing beyond being able to read code
- Tier 3 assumes you know what the old workaround was working around.
 
## How to claim a task
 
1. Read [CONTRIBUTING.md](https://github.com/boostorg/graph/blob/develop/CONTRIBUTING.md) first: it has the superproject setup, the build and test commands, the naming conventions, and the full `BOOST_*` macro policy table that several tasks below refer to.
2. Comment on this issue saying which box you are taking, e.g. *"Taking 2.1 on `strong_components.hpp`"*. We will tick it here so nobody duplicates work.
3. Ask questions in the comments. There is no such thing as a question too basic on this issue.

## Ground rules
 
- **One file per PR.** Every checkbox below is exactly one file and exactly one PR. This is deliberate: small PRs get reviewed much faster, they are easy to revert if we got something wrong, and they let many people contribute in parallel without stepping on each other. Please do not batch several files together to "save time" as it costs review time. The one exception is a header and its matching test when the change genuinely spans both.
- **One task per PR.** Several files appear under more than one task. If your file needs work from two tasks, that is two PRs. Do not mix a macro cleanup with a rename or a formatting pass.
- **No behavior change.** Everything below should be a no-op at runtime. If you find a bug along the way, that is great: open a separate PR, with a regression test.
- **The test suite must stay green.** `./b2` from `libs/graph/test` runs everything (~10 min). For a single test: `./b2 cycle_canceling_test`. If you struggle with local builds, feel free to run on CI.
- **The CI matrix is gcc-14 + clang-19 over C++14/17/20/23.** C++14 is the floor: no structured bindings, no `if constexpr`, no inline variables, no single-argument `static_assert`.
- Do not run `clang-format`. We will soon make a repo-wide formatting PR and integrate `clang-format` in CI.

## Do not touch these files
 
The SGB (Stanford GraphBase) and LEDA adaptors are only built when `SDB` / `LEDA` roots are passed to b2, so **nothing in this list is currently verified by CI**.
 
```
include/boost/graph/leda_graph.hpp
[...]

PR #596, opened 2026-09-20
chore: replace NULL with nullptr in file dimacs.hpp

cxljs:
### Before submitting

### Type of change

- [ ] Bug fix
- [ ] Documentation
- [ ] Build, CI, or tooling
- [ ] Other (specify below)

- [x] No

### What this PR does

replace NULL with nullptr in file dimacs.hpp

### Checklist

- [ ] New behavior is covered by a test, or this is a docs / build / refactor change.


cxljs:
Hi, according to the issue #593 requirements, clang-format should not be run, and each checkbox should correspond to exactly one file and one PR. Therefore, I only modified a single file and did not run clang-format.

If you would prefer to have all NULL usages replaced with nullptr in this PR, I would be happy to update it accordingly.

jeremy-murphy (approved):
Thanks!

Becheler (approved):

Becheler:
Hi @cxljs ! 
Thank you for your contribution ! It looks perfect, I just manually triggered CI. 
Sorry for the small delay I was wandering the Rocky Mountain National Park ! 🤗

cxljs:
> I was wandering the Rocky Mountain National Park ! 🤗

Hi @Becheler , hope you had a great time at Rocky Mountain National Park! 🤗

I actually came across your [CppCon talk](https://cppcon2026.sched.com/event/2WOel/open-content-the-barrier-is-a-myth-becoming-a-boost-maintainer), and that’s what motivated me to take a look at the Boost.Graph project and see if there was anything I could help with. I’d like to contribute something back to the open-source community, while also improving my C++ skills along the way.

Over the weekend, I looked through the code for a few of Graph’s data structures, but I haven’t yet had the chance to understand them deeply or think about whether there are any areas that could be optimized. I’ll keep digging into them and thinking about it. Also, if there are any tasks or areas in Graph that you’d like to work on but haven’t had the time for, feel free to @ me anytime.

On the C++ side, I’ve also been thinking about whether “improving my C++ skills” is becoming a bit of an old-fashioned goal. With how quickly LLMs have developed this year, people are writing less code themselves, and generated code is already good enough for many projects. But I’m still want to improve my C++ skills mainly because I genuinely enjoy getting better at C++ — it’s fun in its own way.

Becheler:
Hi @cxljs !

Yes the National Park was amazing, I do recommend visiting if you can. There is a bus line from Denvers to the park, and then free shuttles to bring you to the trails. It was very convenient and very practical !

> I looked through the code for a few of Graph’s data structures, but I haven’t yet had the chance to understand them deeply or think about whether there are any areas that could be optimized

Yes those data structure code is rather complex and very "load-bearing" as would an AI say. It's a cornerstone of the library so historically people avoided touching those files. If you are interested by those data structure, maybe my meetup presentation could help you getting an idea of what each data structure is doing: https://youtu.be/-OVvzRFiYLU?t=2337

> I’ve also been thinking about whether “improving my C++ skills” is becoming a bit of an old-fashioned goal

If you don't want to be replaced by an AI in the next few years, then I would say it's actually more critical than ever. You don't need to write/generate code as much as before, but it means you need even more understanding of architecture, design choices, refactoring techniques, C++ abstractions (policy classes, traits, SFINAE, concepts etc) to be able to understand what the AI is recommending. 

For example few days ago in a different Boost lib that was failing on msvc, Claude recommended moving an entire feature to a different namespace. I know enough about architecture to know it was a (very) bad idea. Digging a bit more, bringing an operator into local scope with an alias was just enough. If you/we don't do the extra effort of developing skills instead of delegating skills then yes we will become old-fashioned 😄 

In a few words, never ever push a generated code you don't fully understand. And resist the urge to push as much as possible. Projects with dozen of PRs per day are impossible to maintain, review, or ensure code ownership, and are probably doomed to instability by code duplication. It may not be possible to do so in a company that pushes you to send features before tomorrow, but in long-term OSS and critical softwares stability over whims is definitely desirable 😄

